PR #106, opened 2012-01-12
new junit logger and stepToClearField added

rlgomes:
I've added a junit logger and even the ability to plug your own logger into the KIF reporting and I also added a new method called stepToClearField which makes it easy to delete the text in a textfield/textview as a user would (to clear previous state that some apps like to save). If you guys find it useful then feel free to pull the changes in and if you want the logger business to be written slightly different let me know as I'm not an experienced objective c developer and have done things in the easiest and quickest way.

### DIFF
--- a/Additions/UIView-KIFAdditions.m
+++ b/Additions/UIView-KIFAdditions.m
@@ -322,7 +322,6 @@
         return;
     }
     UITouch *touchDown = [[UITouch alloc] initAtPoint:points[0] inView:self];
-    [NSThread sleepForTimeInterval:0.1];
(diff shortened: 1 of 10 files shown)


mikelupo:
@rlgomes So now that KIF is at version 3.2, it seems that there is no clean way to merge this pull request cleanly as the class h/m files have changed pretty drastically.
Any takers?

phatmann:
I will re-open so we can gauge interest.

rlgomes:
I haven't touch any IOS stuff in years so if there are others with interest then do what you must with the changes.

phatmann:
Closing due to lack of action.

Issue #439, opened 2014-07-23
Crash in waitForAccessibilityElement


tapwork:
Hi 

we have a crash (EXC_BAD_ACCESS) in 

``` objective-c
- (UIView *)waitForViewWithAccessibilityLabel:(NSString *)label value:(NSString *)value traits:(UIAccessibilityTraits)traits tappable:(BOOL)mustBeTappable
{
    UIView *view = nil;
    [self waitForAccessibilityElement:NULL view:&view withLabel:label value:value traits:traits tappable:mustBeTappable];
    return view;
}
```

I tried malloc_history and NSZombiesEnabled but could not identify an object with a dangling pointer.
I stepped through the code, because I assume that something odd is going on with the autoreleased out parameter `&view` (view object passed by reference).

I found out that KIF returns a  `UIAccessibilityElementMockView` object when the crash occurs. 
The KIF code has a comment about that:

``` objective-c
- (UIAccessibilityElement *)accessibilityElementWithLabel:(NSString *)label accessibilityValue:(NSString *)value traits:(UIAccessibilityTraits)traits;
{
....
// Check the subviews first. Even if the receiver says it's an accessibility container,
    // the returned objects are UIAccessibilityElementMockViews (which aren't actually views)
    // rather than the real subviews it contains. We want the real views if possible.
    // UITableViewCell is such an offender.
...
}
```

I assume that UIKit is mocking a UIView / UIAccessibilityElement which does not exist in the hierarchy.  This mocked object will be autoreleased, which leads to the crash.
Did you encounter this crash?
We are using KIF 3.0.7

PR #447, opened 2014-08-19
Fixes issue with overreleased objects

tapwork:
The autoreleasepool causes overrelease issues with the out parameters  (NSError *\* and UIView **)
Also I do not see any memory benefits using the autorelease pool here. (tested the footprint)
Out parameters will be implicitly autoreleased. i.e. NSError *__autoreleasing *
See more: http://blog.pioneeringsoftware.co.uk/2012/03/06/out-parameters-when-arcing/
It also fixes the issue we filed here https://github.com/kif-framework/KIF/issues/439


XPCD:
I'm still facing the memory release issues on KIF release 3.2.3
Thought you ought to know.

phatmann:
@XPCD Thanks for letting me know. If you can provide some more details and open a new issue that would be helpful.

PR #480, opened 2014-10-15
iOS 8 fix: UITableViews are now wrapped in a scrollable UITableViewWrapper view

glentregoning:
This causes issues when KIF moves the content of the scroll view instead of the actual UIScrollView dependent, as none of the UIScrollViewDelegate calls are made to the actual instance. 

Excluding this class works around this issue.


glentregoning:
@phatmann this issue is fixed in the 3.1.0 release.

PR #510, opened 2014-12-01
Add support for dismissing system alerts

joemasilotti:
This is a first pass at dismissing system alerts (eg. location services and access to photos). It works by linking to the private UIAutomation framework and then calling down the `UIAElement` chain to grab the alert. It then taps the last button which should be the _Allow_ one.

Also of note is how the private headers are pulled in. In its current state all the knowledge lives in `KIFSystemAlertHandler.m` and only exposes the _required_ methods. Take a peek at the [full header dump](https://github.com/joemasilotti/UIAutomation) and [the inspiration for this](https://github.com/TestingWithFrank/PublicAutomation) if you are interested in seeing the full list. There is also the option of pulling one of those in as a Cocoapod and keeping all the knowledge contained there.

This could be a solution for PR #507 running on CI but I want to first make sure this is an acceptable approach before continuing. If so, please let me know and I will continue with adding support for iOS 7.X and not failing the test if the alert never shows up.

While this will not yet allow interaction with external views (eg. `ABPeoplePickerNavigationController`), it does mark a huge step forward for KIF being able to interact with elements outside of the standard accessibility hierarchy.


alexmanzer:
One question to this feature: Will this handling of system alertviews also tap automatically on "iOS update available"-alertviews like this: 
iOS update alertview

fcy:
I'd love to have a way to dismiss the update alert. Thanks for this awesome use of UIAutomation!

iMaximus:
It not work on IOS 8.4 for location alert :(

phatmann:
@iMaximus can you file a new issue with more details? Thanks!

iMaximus:
Sorry, my bad. I thought that not work method [tester acknowledgeSystemAlert]; for location alert on simulator iphone 6 (8.4) but all work correctly.

tom-wolters:
No longer working with Xcode 8.1/Swift 3? Method dissapeared for me after updating to Xcode 8.1.

phatmann:
@tom-wolters Which method?

tom-wolters:
@phatmann I apologize - The method ('acknowledgeSystemAlert()') is working as intended! I removed my comment.

maryamaffinityclick:
Im using KIF in swift , I cannot have the method acknowledgeSystemAlert(), how can I call this method ??

justinseanmartin:
It is available on `-[KIFUIViewTestActor acknowledgeSystemAlert]` (and `KIFUITestActor` as well).  You should be able to call it just like you call any other action on `tester` or `viewTester` in Swift.

maryamaffinityclick:
right , My target was device so I wasn't able to see the function! 
did you find any solution for devices ?

justinseanmartin:
Gotcha, we do all of our testing on iOS Simulators, so we don't have this issue. 

I believe we can't load the UIAutomation.framework on devices in order to interact with system alerts unless you have a jailbroken device, at which point it should be theoretically possible (though I haven't tried). Rather than relying on these methods to interact with system alerts, you should be able to swizzle some methods to stub the behavior that ends up popping the system alerts.

I posted in #979 with our workaround for how to suppress the CLocationManager alerts. We do something similar for the photo picker as well.

maryamaffinityclick:
You are totally right! 
Although I am thinking of having a single UITest class which will only interact with permission dialogs and run before KIF UnitTests to prepare the app. cause I don't wanna stub the behaviour I want to actually access the photos or location. 
I dont know how it will work in practice but theoretically would works fine and it will run just once on fresh install .

mikelupo:
For those damn update pop ups we run our WiFi connections through a proxy and blacklist the apple sites. We use privoxy.

justinseanmartin:
Regarding not wanting to stub the behavior of accessing the photos or location, at least philosophically for me, it isn't testing behavior that is within your control.

As long as you're conforming to the same API contract that is exposed by Apple's API, it shouldn't be necessary to actually test the iOS system behavior IMO.

hejun-lyne:
Hi, I found that applicationState is always `UIApplicationStateInactive` after called `acknowledgeSystemAlert`, and `UIApplicationDidBecomeActiveNotification | UIApplicationWillResignActiveNotification` will never received.

PR #543, opened 2015-01-07
Add autoreleasepool to prevent memory peak

TBonnin:
Iterating over UICollectionView cells could lead to an accumulation of
cells object (and therefore an increase of memory).
This commit adds an autoreleasepool to mitigate the problem and force
the release of the cell objects.


phatmann:
Good idea. Thanks for this.

PR #546, opened 2015-01-09
Fixes compilation error in Swift example project.

mitochondrion:
Added missing optional chaining.


bnickel:
You'd think they'd keep interop bindings fixed at least in the same minor release. :( lgtm

PR #547, opened 2015-01-12
Adding the property "animationWaitingTimeout" that controls how long tes...

### LINKED ISSUE #530: Performance slow.
On some of my modals views it takes > 1 second to tap something. Any clues on how to debug this?

Sipeso:
...ts will wait for animations to finish. Its default-value is 0.5 second. It can also be deactivated by setting it to 0. This should fix #530.

Also moving the waitForTimeInterval-call into the animationWaitingTimeout-call. The call is necessary to give the views a chance to start animations, before we wait for animations to finish. Thus, the call is necessary that animationWaitingTimeout will work properly.

### DIFF
--- a/Classes/KIFTestActor.m
+++ b/Classes/KIFTestActor.m
@@ -64,6 +64,7 @@
         _line = line;
         _delegate = delegate;
         _executionBlockTimeout = [[self class] defaultTimeout];
+        _animationWaitingTimeout = 0.5f;
     }
     return self;
 }
@@ -179,7 +180,13 @@ static NSTimeInterval KIFTestStepDelay = 0.1;
 }
 
 -(void)waitForAnimationsToFinish {
-    NSTimeInterval maximumWaitingTimeInterval = 3;
+    NSTimeInterval maximumWaitingTimeInterval = self.animationWaitingTimeout;
+    if(maximumWaitingTimeInterval <= 0) {
+        return;
+    }
+    
+    // Wait for the view to stabilize and give them a chance to start animations before we wait for them.
+    [self waitForTimeInterval:0.5f];
     
     NSTimeInterval startTime = [NSDate timeIntervalSinceReferenceDate];
     [self runBlock:^KIFTestStepResult(NSError **error) {
--- a/Classes/KIFTestActor.h
+++ b/Classes/KIFTestActor.h
@@ -89,6 +89,7 @@ typedef void (^KIFTestCompletionBlock)(KIFTestStepResult result, NSError *error)
 @property (nonatomic, readonly) NSInteger line;
 @property (weak, nonatomic, readonly) id<KIFTestActorDelegate> delegate;
 @property (nonatomic) NSTimeInterval executionBlockTimeout;
+@property (nonatomic) NSTimeInterval animationWaitingTimeout;
 
 - (instancetype)usingTimeout:(NSTimeInterval)executionBlockTimeout;
 
--- a/Classes/KIFUITestActor.m
+++ b/Classes/KIFUITestActor.m
@@ -168,9 +168,6 @@
         
         return KIFTestStepResultSuccess;
     }];
-    
-    // Wait for the view to stabilize.
-    [self waitForTimeInterval:0.5];
 
     [self waitForAnimationsToFinish];
 }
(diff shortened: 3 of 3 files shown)


jacobrelkin:
We pulled these changes into an integration branch, and it turns out that this works great for us! We had quite a few failures due to timing issues introduced by #530, so addressing this is a high priority for us.

Can someone kick off another build on Travis? Looks like there were some simulator issues that prevented a proper test run.

Thanks for the timely fix, @Sipeso! :+1:

phatmann:
I kicked off a Travis build a few hours ago, but it failed again. Will try one more time. If I can get a local build to pass (very likely), I will merge these changes in and push a new release.

phatmann:
The test passes locally so I am going to merge this. Thanks @Sipeso for making this fix!

Classes/KIFTestActor.m
```diff
@@ -179,7 +180,13 @@ - (void)waitForTimeInterval:(NSTimeInterval)timeInterval
...
 
 -(void)waitForAnimationsToFinish {
-    NSTimeInterval maximumWaitingTimeInterval = 3;
+    NSTimeInterval maximumWaitingTimeInterval = self.animationWaitingTimeout;
+    if(maximumWaitingTimeInterval <= 0) {
+        return;
+    }
+    
+    // Wait for the view to stabilize and give them a chance to start animations before we wait for them.
+    [self waitForTimeInterval:0.5f];
```
justinseanmartin: The `Wait for the view to stabilize` half second wait predated your changes, but is now directly tied to using `waitForAnimationsToFinish` with a maximumWaitingTimeInterval > 0. Can we move that above the conditional, or leave that directly in the callers in KIFUITestActor.m?

Classes/KIFTestActor.m
```diff
@@ -179,7 +180,13 @@ - (void)waitForTimeInterval:(NSTimeInterval)timeInterval
 }
 
 -(void)waitForAnimationsToFinish {
```
justinseanmartin: Animations are a UI concept only, so I'd expect this would make more sense living in KIFUITestActor than KIFTestActor.

Classes/KIFTestActor.m
```diff
@@ -179,7 +180,13 @@ - (void)waitForTimeInterval:(NSTimeInterval)timeInterval
...
+    NSTimeInterval maximumWaitingTimeInterval = self.animationWaitingTimeout;
+    if(maximumWaitingTimeInterval <= 0) {
+        return;
+    }
+    
+    // Wait for the view to stabilize and give them a chance to start animations before we wait for them.
+    [self waitForTimeInterval:0.5f];
     
     NSTimeInterval startTime = [NSDate timeIntervalSinceReferenceDate];
     [self runBlock:^KIFTestStepResult(NSError **error) {
```
justinseanmartin: Rather than using `runBlock:timeout:` and passing a timeInterval + 1 second to prevent the exception from being raised, it seems like it would make more sense to use `tryRunningBlock:complete:timeout:error:`. That way, you don't need your own time interval checking on line 209.

Classes/KIFTestActor.m
phatmann: The original 0.5 second wait was to allow animations to complete, so it was really a poor man's `waitForAnimationsToFinish`. Now that we have a real function to do this, we do not need the callers in KIFUITestActor.m to manually do the wait anymore. Furthermore, if a client sets the `maximumWaitingTimeInterval` to 0, this allows them to skip the whole stabilization delay, which is a good thing. On the other hand, you did find a bug: if the `maximumWaitingTimeInterval` is less than 0.5 seconds, then we should call waitForTimeInterval using the maximum and then get out. I filed issue #550 to cover this.

Classes/KIFTestActor.m
phatmann: Great point. Fixed in 73e27dfd0.

Classes/KIFTestActor.m
phatmann: @Sipeso, what do you think?

Classes/KIFTestActor.m
justinseanmartin: Thanks for addressing!

Classes/KIFTestActor.m
justinseanmartin: This is more style, so it obviously isn't absolutely necessary. It just makes the code a bit easier to read and the logic a bit simpler in my opinion. Maybe something for @Sipeso to keep in mind in the future.

PR #549, opened 2015-01-13
Fix documentation typos

RoyalPineapple:
elemetn -> element


phatmann:
Thanks, I like to catch these!

PR #554, opened 2015-01-22
Add waitForAnimationsToFinishWithTimeout method

TBonnin:
Allows to specify a waiting timeout without having to modify the global animationWaitingTimeout


TBonnin:
I am working on a screen which has a long artistic animation (> 3s). I would like to be able to specify the amount of time the KIFTestActor wait for the animation to finish without having to modify the global animationWaitingTimeout.

What do you think?

phatmann:
Good little change, thanks.

PR #564, opened 2015-02-03
Add method for remove the row in a table view

zixzelz:


joemasilotti:
@zixzelz would it be possible to add a test for the new method? There are a lot of examples in [TableViewTests](https://github.com/kif-framework/KIF/blob/master/KIF%20Tests/TableViewTests.m) that you could use as a base.

phatmann:
I like this change, and I like the way you refactored the swipe method to avoid copying code. But I would like to see a test for it before I merge this PR. Any chance you could code one up?

zixzelz:
Yes, I think I can create test for that on weekend, when i will be have more free time.

phatmann:
@zixzelz any progress on this?

justinseanmartin:
No updates since 2015, going to close this. Would be happy to revisit if someone wants to revive this branch.

PR #570, opened 2015-02-13
Add workaround for error finding XCTest.h

bradfeehan:
Similar fix to kif-framework#420, but when being used without CocoaPods (e.g. Git submodules).


phatmann:
Glad to see this fixed, thanks.

Issue #555, opened 2015-01-23
KIFUITestActor closely couples finding a screen element with acting upon it. 


RoyalPineapple:
We currently have 5 methods used to clear text from a text view. None of these support matching an element by a predicate, or by block, or any other trait.
`- (void)clearTextFromFirstResponder;`
`- (void)clearTextFromViewWithAccessibilityLabel:(NSString *)label;`
`- (void)clearTextFromViewWithAccessibilityLabel:(NSString *)label traits:(UIAccessibilityTraits)traits;`
`- (void)clearTextFromElement:(UIAccessibilityElement *)element inView:(UIView*)view;`
`- (void)clearTextFromViewWithAccessibilityIdentifier:(NSString *)accessibilityIdentifier;`

This pattern exists for every interaction we might take. `KIFUITestActor` currently contains 69 action methods (with another 10 for conditionals, and 13 for identifiers) the majority of which are similarly duplicated.

`KIFUITestActor` could be replaced with a new actor which would be instantiated with methods used for finding any element.
For example:
`- (instancetype)usingAccessibilityLabel:(NSString *)label;`
`- (instancetype)usingAccessibilityIdentifier:(NSString *)identifier;`
`- (instancetype)usingPredicateWithFormat:(NSString *)predicateFormat, ..`

Combined with methods used for acting upon the element.
`- (void)tap;`
`- (void)longPress;`
`- (void)longPressWithDuration:(NSTimeInterval)duration;`

This would simplify the API to 4 or 5 methods used for finding elements and about 20 methods for possible interactions. 

The explicit intent would be to avoid makeing any breaking changes but to expose a new API allowing the same functionality. This would also allow addition of new actions or matching methods without the need to add all the permutations. For example we were considering adding matching based on block evaluation which would be much less tedious with this new design. 

We wanted to float this idea before heading too far down the rabbit hole. We've done an initial proof of concept which we would love to share if the idea is well recieved by the KIF community. 
Please provide feedback, thoughts, and suggestions. 

Our concept actor is derived from `KIFTestActor` and internally leverages an instance of `KIFUITestActor` to avoid duplicating code. If our proposal is well recieved we could eventually move that code into our actor while deprecating the existing `KIFUITestActor`.

phatmann:
I would love to see your work here! The proliferation of methods has started to bother me as of late, so your timing is perfect.

RoyalPineapple:
Glad you're interested! 
[...]

PR #579, opened 2015-02-19
KIFUIViewTestActor

RoyalPineapple:
New KIFUIViewTestActor as outlined in https://github.com/kif-framework/KIF/issues/555
This actor decouples finding an element from acting upon the element.

This also allows you to define an actor for a specific view description, allowing you to reference the same view multiple times without worrying about stringly-typed errors. 
(see https://github.com/RoyalPineapple/KIF/blob/ca1b2baa5189e133c6aa50c6cbd86a1857f606c0/KIF%20Tests/PickerTests_ViewTestActor.m) for an example

This change will increase maintainiblity of the kif framework as well as our individual tests.

### DIFF
--- a/Classes/KIFUIViewTestActor.m
+++ b/Classes/KIFUIViewTestActor.m
@@ -1,0 +1,395 @@
+//
+//  KIFUIViewActor.m
+//  KIF
+//
+//  Created by Alex Odawa on 1/21/15.
+//
+//
+
+#import "KIFUIViewTestActor.h"
+#import "UIWindow-KIFAdditions.h"
+
+
+@interface KIFUIViewTestActor ()
+
+@property (nonatomic, strong, readonly) KIFUITestActor *actor;
+@property (nonatomic, strong, readwrite) NSPredicate *predicate;
+
+@end
+
+
+@implementation KIFUIViewTestActor
+
+#pragma mark - Initialization
+
+- (instancetype)usingPredicate:(NSPredicate *)predicate;
+{
+
+    [self _appendPredicate:predicate];
+    return  self;
+}
+
+- (instancetype)usingAccessibilityLabel:(NSString *)accessibilityLabel;
+{
+    int systemVersion = [UIDevice currentDevice].systemVersion.intValue;
+
+    if ([accessibilityLabel rangeOfString:@"\n"].location == NSNotFound || systemVersion == 6) {
+        return [self usingPredicate:[NSPredicate predicateWithFormat:@"accessibilityLabel == %@", accessibilityLabel]];
+    }
+
+    // On iOS 6 the accessibility label may contain line breaks, so when trying to find the
+    // element, these line breaks are necessary. But on iOS 7 the system replaces them with
+    // spaces. So the same test breaks on either iOS 6 or iOS 7. iOS 8 befuddles this again by
+    // limiting replacement to spaces in between strings.
+    // UNLESS the accessibility label is set programatically in which case the line breaks remain regardless of OS version.
+    // To work around this replace the line breaks using the preferred method and try matching both.
+
+    //this feels horribly hacky and looks bad in our - (NSString *)description :( but tests are passing
+    // We might consider replaceing the predicate with a block that can do cleaner checking,
+
+    NSString *alternate = nil;
+    if (systemVersion == 7) {
+        alternate = [accessibilityLabel stringByReplacingOccurrencesOfString:@"\n" withString:@" "];
+    } else {
+        alternate = [accessibilityLabel stringByReplacingOccurrencesOfString:@"\\b\\n\\b" withString:@" " options:NSRegularExpressionSearch range:NSMakeRange(0, accessibilityLabel.length)];
+    }
+
+    return [self usingPredicate:[NSPredicate predicateWithFormat:@"accessibilityLabel == %@ OR accessibilityLabel == %@", accessibilityLabel, alternate]];
+}
+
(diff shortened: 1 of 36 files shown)


Classes/KIFUIViewTestActor.m
```diff
@@ -0,0 +1,396 @@
...
+}
+
+- (instancetype)usingValue:(NSString *)value;
+{
+    return [self usingPredicate:[NSPredicate predicateWithFormat:@"accessibilityValue MATCHES %@", value]];
+}
+
+- (instancetype)usingExpectedClass:(Class)expectedClass;
+{
+    return [self usingPredicate:[NSPredicate predicateWithFormat:@"class MATCHES %@", expectedClass]];
```
TBonnin: Should not it be 
```
return [self usingPredicate:[NSPredicate predicateWithBlock:^BOOL(id evaluatedObject, NSDictionary *bindings) {
    return [evaluatedObject isKindOfClass:expectedClass];
}]];
```

Classes/KIFUIViewTestActor.m
```diff
@@ -0,0 +1,396 @@
...
+    [self _appendPredicate:predicate];
+    return  self;
+}
+
+- (instancetype)usingAccessibilityLabel:(NSString *)accessibilityLabel;
+{
+    int systemVersion = [UIDevice currentDevice].systemVersion.intValue;
+
+    if ([accessibilityLabel rangeOfString:@"\n"].location == NSNotFound || systemVersion == 6) {
+        return [self usingPredicate:[NSPredicate predicateWithFormat:@"accessibilityLabel MATCHES %@", accessibilityLabel]];
```
TBonnin: Using MATCHES implies the right hand expression is a regex. Why not use '==' instead? The right hand expression would always be a accessibilityLabel, accessibilityIdentifier, traits or value.

Classes/KIFUIViewTestActor.m
RoyalPineapple: Block predicates create an issue with our error logging as their `description` isn't very usefull.  You're absolutely right though, this is incorrect and wont work with subclasses, etc.

Thinking about it more, I dont think we should support this at all. I had initially implemented it to debug an issue where KIF was finding the wrong element, but the whole concept of a class is an implementation detail and  isn't user-facing, so it really doesnt belong. If needed anyone can build a class predicate and pass it into `usingPredicate:`

Classes/KIFUIViewTestActor.m
RoyalPineapple: You are correct, this was an oversignt.

phatmann:
@RoyalPineapple, sorry to not respond sooner to this. I had a backlog of pull requests with crucial bug fixes I wanted to get through first.

As I said before, I like this change. Let me review it and think about the best way to roll it out. My thought right now is to place it in a separate subspec so people could review it, try it out, and tweak as needed. After that the next step would be to decide whether or not to deprecate the current API in favor of this one, which would then be the birth of KIF 4.0.

I encourage everyone to look at this pull request and weigh in with your thoughts and concerns.

In the meantime, @RoyalPineapple, can you please squash the 3 commits?

RoyalPineapple:
@phatmann I've squashed the commits and rebased off the upstream master. 
I also updated the UIViewTestActor tests to use category methods for commonly referenced views.

steipete:
As someone who just recently started to look into KIF, this PR is VERY interesting and a huge step forward. Definitely justifies a KIF 4.0 tag.

phatmann:
In order to merge this excellent pull request I want to gather more feedback and figure out the best way to package it up. I will be very busy until April 15 or so, so please be patient. You can be sure this is not being neglected! 

In the meantime, if anyone comes forward with a concrete packaging suggestion and is willing to implement it, I am all ears :-)

phatmann:
I am back now, and I will make a proposal for how to package this up this week or next week.

steipete:
Great! We've been adding tons of KIF tests so far and are happy to test a v4 branch.

mokagio:
I'd be happy to test version 4 too 👍

phatmann:
Lately there has been more and more confusion about the Identifier tests and the Conditional tests since they are separated out. In addition lots of folks want a new kind of test that works with either `accessibilityLabel` or `accessibilityIdentifier`. To avoid a combinatorial explosion, I believe we need to use a solution like this pull request. Let me review this pull request again, see if it feels good and suggest any possible changes. Once we have a plan we all agree on, we can go ahead and create KIF 4.0 based on it.

No matter what we decide, we need to make sure we have 100% backward compatibility with KIF 3.0 via a "legacy" layer.

steipete:
Fully agree that the current API is a bit confusing here. How about you create a new v4-branch based on this PR and then we figure out the details as we migrate the API to the new version. I don't really care about legacy support, that just clutters up autocomplete and makes the API larger.

Can I help out here? KIF's becoming really important for pspdfkit.com 's testing, so happy to put some time in.

RoyalPineapple:
Backwards compatability is important to me as we have a lot of tests that have yet to be updated to the new actor. 
As long as we keep the actors as seperate classes we should be able to move forward incrementally without polluting the autocomplete.

phatmann:
@RoyalPineapple, the idea of separate legacy classes using the existing names is a great idea. It will make backwards compatibility automatic for existing users, yet have no negative impact for new users.

phatmann:
I created a new v4 branch and merged this pull request into it. Let's move the discussion to #646.

Issue #574, opened 2015-02-16
Incompatible with Xcode 6.3 / iOS 8.3 (EXC_BAD_ACCESS)


siemensikkema:
The existing test suite breaks under Xcode 6.3. There is an EXC_BAD_ACCESS at line 394 in UIView-KIFAdditions.m as a result of calling ```tapViewWithAccessibilityLabel:```. It seems something changed under the hood related to UITouch.
See also #562, specifically https://github.com/kif-framework/KIF/issues/562#issuecomment-74267256

gurgeous:
+1

a-dogan:
Seeing this as well, same version of Xcode. In my case using [tester waitForViewWithAccessibilityLabel:@"Cancel"]; works, but when I try to tap it I get the same exception as you.

Twyly:
+1

codeman9:
+1

triage:
+1

TBonnin:
To give a little bit more details. I just poked around a few minutes and this is what I can report:
`[UIView TapAtPoint:]` indeed crashes at line https://github.com/kif-framework/KIF/blob/a98d41/Additions/UIView-KIFAdditions.m#L402 with following error `(EXC_BAD_ACCESS, code=1, address=0x20)`

I suspect the UIEvent internals to have changed so the UIEvent object constructed by the `[UIView eventWithTouches:]` method https://github.com/kif-framework/KIF/blob/a98d41/Additions/UIView-KIFAdditions.m#L756 via calls to private API is probably incorrect.

rvangraan:
We get the same crash too. Any suggestions on how to fix it?

phatmann:
I have not tested with the 6.3 beta yet. Has anyone dumped the private headers for clues as to what has changed?

TBonnin:
Hi,
I have a tentative fix for this issue. See https://github.com/kif-framework/KIF/pull/582
I have tested on iOS 7.1, 8.1 and 8.3 but I had not a lot of devices handy. I would appreciate a lot if anybody could test my patch on more devices/OS.

Thank you

Tulkas:
I have the same issue, but I get the error in the line 180 of KIFUITestActor.

phatmann:
Fixed by #582

luoxubin:
[...]

PR #581, opened 2015-02-27
Possibly fix for XCode 6.3 sendEvent BAD_ACCESS 

sebbean:
replaced UIApplication sendEvent: with touchesBegan/touchesEnded

### DIFF
--- a/Additions/UITableView-KIFAdditions.m
+++ b/Additions/UITableView-KIFAdditions.m
@@ -47,7 +47,7 @@
     [touch setPhaseAndUpdateTimestamp:UITouchPhaseBegan];
     
     UIEvent *eventDown = [self eventWithTouch:touch];
-    [[UIApplication sharedApplication] sendEvent:eventDown];
+    [self touchesBegan:eventDown.allTouches withEvent:eventDown];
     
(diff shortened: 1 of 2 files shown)


phatmann:
@sebbean, can you please check in the test of the changes?

I would feel more comfortable with this PR if we could find an explanation of what is causing the #574 crash, so we can verify that this solution makes sense. If I have time I will dig in on this.

phatmann:
I am also looking at #582 to see which solution seems better.

phatmann:
I took a look at #582. At least we now understand the cause of the crash. #582 solves the issue but requires a lot of extra code. The solution proposed in this PR skirts the whole issue by avoiding sending events, but it means that KIF is not actually simulating taps any more. I am afraid that this will break people's tests, but perhaps someone could convince me otherwise.

sebbean:
Note: this fix does not work for swipes or long presses

phatmann:
@sebbean I misunderstood your earlier comment, I thought you had fixes for swipes and long presses on your machine. If you do not, then we need to go with #582. I am going to close out this PR. Please re-open if you find way to make a complete fix and we have not yet merged #582.

PR #582, opened 2015-03-02
Make touch UIEvent compatible with iOS 8.3

TBonnin:
Tentative fix for https://github.com/kif-framework/KIF/issues/574

Running KIF on iOS 8.3 (with XCode 6.3 Beta) leads to the following error
screen shot 2015-03-01 at 10 51 33 pm

Since it was referencing `IOHIDEvent` I looked into `[UIEvent _hidEvent]` [private attribute](https://github.com/nst/iOS-Runtime-Headers/blob/master/Frameworks/UIKit.framework/UIEvent.h#L15) and realized it was always nil for [UIEvent created by KIF](https://github.com/kif-framework/KIF/blob/a98d4102e6ee1f11e7fc88bf99ab724baf320ce5/Additions/UIView-KIFAdditions.m#L782) (unlike real UIEvents that have a valid `_hidEvent`).
My guess was that starting with iOS 8.3, the usage of `[UIEvent _gsEvent]` was deprecated in favor on `[UIEvent _hidEvent]`.
After some poking around, trial and error and the help of [some code](https://github.com/iolate/SimulateTouch/blob/master/SimulateTouch.mm#L170) found on github I was able to fake this `IOHIDEvent` attribute.
This solution relies heavily on iOS private APIs (IOKit private framework needs to be added) so I am not sure this is the best approach. @phatmann what do you think?

I have tested on iOS 7.1, 8.1 and 8.3.

### DIFF
--- a/Additions/UIEvent+KIFAdditions.h
+++ b/Additions/UIEvent+KIFAdditions.h
@@ -1,0 +1,19 @@
+//
+//  UIEvent+KIFAdditions.h
+//  KIF
+//
+//  Created by Thomas on 3/1/15.
+//
+//
+
+#import <UIKit/UIKit.h>
+
+// Exposes methods of UITouchesEvent so that the compiler doesn't complain
+@interface UIEvent (KIFAdditionsPrivateHeaders)
+- (void)_addTouch:(UITouch *)touch forDelayedDelivery:(BOOL)arg2;
+- (void)_clearTouches;
+@end
+
+@interface UIEvent (KIFAdditions)
+- (void)kif_setEventWithTouches:(NSArray *)touches;
+@end
--- a/Additions/UIEvent+KIFAdditions.m
+++ b/Additions/UIEvent+KIFAdditions.m
@@ -1,0 +1,152 @@
+//
+//  UIEvent+KIFAdditions.m
+//  KIF
+//
+//  Created by Thomas on 3/1/15.
+//
+//
+
+#import "UIEvent+KIFAdditions.h"
+#import "LoadableCategory.h"
+#import <mach/mach_time.h>
+
+MAKE_CATEGORIES_LOADABLE(UIEvent_KIFAdditions)
+
+/* IOKit Private Headers */
+#ifdef __LP64__
+typedef double IOHIDFloat;
+#else
+typedef float IOHIDFloat;
+#endif
+typedef struct __IOHIDEvent * IOHIDEventRef;
+typedef UInt32 IOOptionBits;
+typedef uint32_t IOHIDDigitizerTransducerType;
+void IOHIDEventAppendEvent(IOHIDEventRef event, IOHIDEventRef childEvent);
+enum {
+    kIOHIDDigitizerTransducerTypeStylus  = 0x20,
+    kIOHIDDigitizerTransducerTypePuck,
+    kIOHIDDigitizerTransducerTypeFinger,
+    kIOHIDDigitizerTransducerTypeHand
+};
+enum {
+    kIOHIDDigitizerEventRange                               = 0x00000001,
+    kIOHIDDigitizerEventTouch                               = 0x00000002,
+    kIOHIDDigitizerEventPosition                            = 0x00000004,
+    kIOHIDDigitizerEventStop                                = 0x00000008,
+    kIOHIDDigitizerEventPeak                                = 0x00000010,
+    kIOHIDDigitizerEventIdentity                            = 0x00000020,
+    kIOHIDDigitizerEventAttribute                           = 0x00000040,
+    kIOHIDDigitizerEventCancel                              = 0x00000080,
+    kIOHIDDigitizerEventStart                               = 0x00000100,
+    kIOHIDDigitizerEventResting                             = 0x00000200,
+    kIOHIDDigitizerEventSwipeUp                             = 0x01000000,
+    kIOHIDDigitizerEventSwipeDown                           = 0x02000000,
+    kIOHIDDigitizerEventSwipeLeft                           = 0x04000000,
+    kIOHIDDigitizerEventSwipeRight                          = 0x08000000,
+    kIOHIDDigitizerEventSwipeMask                           = 0xFF000000,
+};
+IOHIDEventRef IOHIDEventCreateDigitizerEvent(CFAllocatorRef allocator, AbsoluteTime timeStamp, IOHIDDigitizerTransducerType type,
+                                             uint32_t index, uint32_t identity, uint32_t eventMask, uint32_t buttonMask,
+                                             IOHIDFloat x, IOHIDFloat y, IOHIDFloat z, IOHIDFloat tipPressure, IOHIDFloat barrelPressure,
+                                             Boolean range, Boolean touch, IOOptionBits options);
+IOHIDEventRef IOHIDEventCreateDigitizerFingerEventWithQuality(CFAllocatorRef allocator, AbsoluteTime timeStamp,
+                                                              uint32_t index, uint32_t identity, uint32_t eventMask,
+                                                              IOHIDFloat x, IOHIDFloat y, IOHIDFloat z, IOHIDFloat tipPressure, IOHIDFloat twist,
+                                                              IOHIDFloat minorRadius, IOHIDFloat majorRadius, IOHIDFloat quality, IOHIDFloat density, IOHIDFloat irregularity,
+                                                              Boolean range, Boolean touch, IOOptionBits options);
+
+/* END of IOKit Private Headers */
+
+//
+// GSEvent is an undeclared object. We don't need to use it ourselves but some
+// Apple APIs (UIScrollView in particular) require the x and y fields to be present.
+//
+@interface KIFGSEventProxy : NSObject
+{
+@public
+    unsigned int flags;
+    unsigned int type;
+    unsigned int ignored1;
+    float x1;
+    float y1;
+    float x2;
+    float y2;
+    unsigned int ignored2[10];
+    unsigned int ignored3[7];
+    float sizeX;
+    float sizeY;
+    float x3;
+    float y3;
+    unsigned int ignored4[3];
+}
+@end
+
+@implementation KIFGSEventProxy
+@end
+
+typedef struct __GSEvent * GSEventRef;
+
+@interface UIEvent (KIFAdditionsMorePrivateHeaders)
+- (void)_setGSEvent:(GSEventRef)event;
+- (void)_setHIDEvent:(IOHIDEventRef)event;
+@end
+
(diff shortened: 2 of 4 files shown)


phatmann:
Nice work on this, @TBonnin! I like this solution because it solves the root cause of the problem. Two questions:

* Do we have to pull in all of IOKit? Can we just copy/paste the code we need into UIEvent+KIFAdditions?

* The Travis tests are failing. Are they passing on your machine? I am re-running them now to see if the problem is intermittent.

phatmann:
@TBonnin how do you feel about the solution proposed in #581?

TBonnin:
I will have a look into this tonight or tomorrow.

phatmann:
@TBonnin, don't worry about the Travis failures. I just ran the tests locally on your branch and the only failure was in PickerTests, which is a known issue (#590).

TBonnin:
@phatmann 
I have removed the unnecessary IOKit headers and only kept the necessary definitions in [UIEvent+KIFAddition](https://github.com/TBonnin/KIF/blob/888f408c18355f2dd4afea9f46132c3d82b6e532/Additions/UIEvent%2BKIFAdditions.m#L15)

phatmann:
@TBonnin, this is great, your solution is now a slam dunk.

One thing: we need to squash your two commits. I recommend that you do it, because then you get clear credit in the repo history when I merge, but if you don't have time I can do it. Let me know.

TBonnin:
commits squashed @phatmann

LeffelMania:
I won't comment on the merit of this particular fix because I don't know the details and I have faith that @TBonnin did some good work here, but adding a dependency on a private Framework (that doesn't seem to be distributed with Xcode even) in a library that I assume most people use as Cocoapod seems... not ideal to me. Anyone updating to HEAD right now is finding a link error in their project, seemingly out of nowhere:

screen shot 2015-03-12 at 12 15 29 am

Which led me here to look at the commit history, where I found this PR, but I'm still not sure how to actually find or link IOKit into my project, as it doesn't show up in the Framework navigator.

screen shot 2015-03-12 at 12 21 24 am

Is there something obvious I'm missing here to get linked properly? And is this a dependency you want to unleash on all of your users in the next release?

TBonnin:
IOKit is distributed with Xcode. Go to `/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/System/Library/Frameworks/` (where all the other Apple Frameworks are located btw) and you will find IOKit among them. 
I don't know why it doesn't show in the 'Add a Framework' dialog.

Note: if you get a `Missing required architecture` warning for all the frameworks then. Go to your target build setting and remove `"$(DEVELOPER_DIR)/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/System/Library/Frameworks"` from the `Framework search path` setting that Xcode automatically added. :/

This definitely needs to be documented in the README the same way the dependency on CoreGraphics is currently documented.

On another note @LeffelMania I agree with you that adding a dependency to the KIF project is not ideal. 
I would be happy to keep exploring alternative solutions to this UIEvent internals change before @phatmann releases the new version.

phatmann:
@LeffelMania, good catch. I updated the Podspec and README to reflect the depedency on IOKit. I tested the change by adding the current version of KIF to a project via CocoaPods, and I did not get any linker errors.

phatmann:
Note that users using `:head` in their Podfile will have linker errors, because they will be using the current code with the older podspec file in the CocoaPods repo. This is a flaw in CocoaPods. The only solution is to push a new release, so I will do that today unless there are strenuous objections that need addressing.

mikelupo:
Whereas I'm still using xcode 6.1.1, I am having this issue and the workarounds listed above do not work for me. Actually, I don't see any path in my xcode for the target that's failing to build in "Framework Search Paths". 
""$(DEVELOPER_DIR)/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/System/Library/Frameworks" from the Framework search path setting that Xcode automatically added. :/"
So now I can't compile using HEAD. 

Question, will this update from @TBonnin fix the swipe gestures that reportedly stopped working in iOS 7?

TBonnin:
What errors do you get @mikelupo

phatmann:
I am going to do the release now, so that CocoaPods folks using `:head` are not stuck. If there are additional fixes needed I will do another release ASAP.

mikelupo:
I get the undefined architecture for i386. Sorry I don't have the exact error any longer.
I can't build it at all now since we are experimenting with trying to convert this KIF distribution into a .framework so it can be pulled in at build time.

mikelupo:
@TBonnin I am still having the linker errors. 
I updated my XCode to 6.2 on my Yosemite.
I started fresh and integrated KIF via via submobule from github. 

Below is my linker error:

Undefined symbols for architecture x86_64:
  "_IOHIDEventAppendEvent", referenced from:
      -[UIEvent(KIFAdditions) kif_setIOHIDEventWithTouches:] in libKIF.a(UIEvent+KIFAdditions.o)
  "_IOHIDEventCreateDigitizerEvent", referenced from:
      -[UIEvent(KIFAdditions) kif_setIOHIDEventWithTouches:] in libKIF.a(UIEvent+KIFAdditions.o)
  "_IOHIDEventCreateDigitizerFingerEventWithQuality", referenced from:
      -[UIEvent(KIFAdditions) kif_setIOHIDEventWithTouches:] in libKIF.a(UIEvent+KIFAdditions.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)

I didn't have the exact line you mentioned in the Framework Search Paths. But what I did have ($(SDKROOT)/Developer/Library/Frameworks) I removed. Didn't make any difference.

Any help is appreciated.

TBonnin:
@mikelupo 
I am assuming you are not using CocoaPods

1. Import the IOKit framework to your project. It is located with the rest of the Apple framework at `/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/System/Library/Frameworks/`
2. Go to your build settings and remove `$(DEVELOPER_DIR)/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk/System/Library/Frameworks` from the `Framework search path` setting

mikelupo:
@TBonnin yes, I am no longer using cocoapods.

What is the proper way to import this into my project? Via Build Phases? Should I do it like this? [apple article](https://developer.apple.com/library/ios/recipes/xcode_help-structure_navigator/articles/Adding_a_Framework.html)

mikelupo:
oops, I see that you stated the same way that I imported coregraphics.
:+1:

TBonnin:
There is a few ways to do so:
1. Add a file like mentioned in your link
2. Drag and drop from the finder
3. Add it from the `Link binary from Library` build phase of your target

mikelupo:
(writing this for future readers, including myself) Weird that I can't traverse down past the xcode.app folder. So what I did was selected another framework in the project explorer and right click and choose, "show in finder". Then I was able to find the IOKit.framework as a sibling. From there I was able to drag it into my Link Binary With Libraries in the Build Phases tab of my target.
This is progress.

So now I went from 6 linker errors to 89. Grrrr. I wonder if that's from my accidently linking in xcode.app? Once I did that, xcode had the beachball spinning for a bit so I force quit. 
I'll have to revert my repo back to HEAD and start cleanly to rule that out now. Damn!

TBonnin:
@mikelupo 
Once you have added the IOKit framework, follow step 2. and remove the unnecessary line in you build setting `Framework search path` that was added by Xcode automatically.

mikelupo:
@TBonnin I had done step 2. I was getting all sorts of crazy linker errors. Once I reverted back to head, reimported my KIF stuff, performed your steps 1 and 2, created a FOO test cases that references kif.h all is well. 
Thanks for your support and your patience.

TBonnin:
Glad you figured it out @mikelupo

Professorq:
After adding IOKit to the 'Link Binary With LIbraries' list, I get a clang crash. Has anyone else seen this and been able to work around it?

clang: error: unable to execute command: Segmentation fault: 11

pyraman:
If you are building a dylib and dynamically inject into an ipa file, then you cannot load the framework this way. So instead of adding to xcode, use dlopen instead. :D.

phatmann:
@pyraman Care to explain exactly what you are are suggesting?

luoxubin:
@TBonnin   @phatmann ，Is there an other way to fake a touch event without rely on IOKit， just  like the way as before 582， not use pure C private APIs. Because my app need to submit to AppStore, and need to simulate touch event  in some  scene.  I have ask many friends and still have not work it out.  Can you help me?
      Many thanks.

pyraman:
I think touchBegan & touchEnded can help you in some cases but they are not
the whole story. You will face long touch prob for ex.
On May 27, 2015 1:38 PM, "罗序斌" <notifications@github.com> wrote:
>
> @TBonnin @phatmann ，Is there an other way to fake a touch event without
rely on IOKit， just like the way as before 582， not use pure C private
APIs. Because my app need to submit to AppStore, and need to simulate touch
event in some scene. I have ask many friends and still have not work it
out. Can you help me?

luoxubin:
I didn't know how to pass the apple API scanning with using a pure C private API. If use Obj-c private API , we can use runtime features to combine the method name dynamically and the performSelector or NSInvocation it.

luoxubin:
@pyraman   Thank you for your  reply. touchBegan & touchEnded  cannot meets my needs. I need to simulate a touch Event ( include set touch point ,touch phase ,tap count and so  on) such as KIF does.

phatmann:
@luoxubin This is not the right forum for this discussion. I recommend trying stackoverflow.

luoxubin:
@phatmann  thanks a lot.  Is there an other way to fake a touch event without rely on IOKit， just like the way as before #582， not use pure C private APIs?

mciarlo:
I appear to be having this issue on Xcode 7.1 running iOS 9.1. I have the same stack and `EXC_BAD_ACCESS` error.
screen shot 2015-11-02 at 10 42 19 am

pyraman:
@phatmann I have never used the library to perform test but I learned something from you.
I guest that the library will be built as a static library and link into iOS project. This requires tester to have a copy of iOS app source code. This is not good.
I'm working on a similar project but we build our project as a dynamic library and perform dynamic link with AUT. With this way, It's no need for tester to have a copy of iOS project source code but an ipa file.

phatmann:
@pyraman So you are saying that your testers can write KIF tests without having a copy of the app sources? An interesting use case. Is this connected in some way to this issue?

PR #588, opened 2015-03-09
Add URL Matcher for asserting on URL scheme

mickeyreiss:
This is useful for making an assertion that an app switches to a
particular app.

It has the benefit of being less technically prescriptive than a fully specified absolute URL,
but still more valuable in tests than a the more general assertion on _any_ URL.


mickeyreiss:
Thanks!

PR #599, opened 2015-03-17
fix typist to work on device with split keyboards

brightredchilli:
This pull request fixes kiftypist behaviour on device with split keyboards. The manifestation was that enterTextIntoFirstResponder will fail when on device with split keyboard, because the UIKeyboardWillShowNotification never gets fired under those circumstances.

### DIFF
--- a/Classes/KIFTypist.m
+++ b/Classes/KIFTypist.m
@@ -53,18 +53,27 @@ static NSTimeInterval keystrokeDelay = 0.01f;
 
 - (void)registerForNotifications
 {
-    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(keyboardDidShowNotification:) name:UIKeyboardDidShowNotification object:nil];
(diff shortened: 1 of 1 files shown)


phatmann:
Looks good, nice work.

brightredchilli:
@phatmann, any chance we can make a tag and cocoapods for this? It's really too bad that there is not an automated way to do this. We use KIF on our project heavily and would benefit from more granular updates. I could do it too but I would need pull access to the repository.

phatmann:
@brightredchili, is there any reason you cannot use `:head` in your Podfile?

In the meantime, let me review the latest commits and see if another release is appropriate. I usually like to resolve all the pull requests "in progress" first.

brightredchilli:
The problem with using 'head' or 'master' is that 'pod update' will move the dependencies of KIF forward, when I really want just 3.2.xx, for example. Somewhat of an annoyance but we can work around it

TBonnin:
I think you can specify a specific commit `:commit => '082f8319af'`

phatmann:
I have just pushed 3.2.1 with this fix in it.

PR #600, opened 2015-03-17
Use iOS default "Build Active Architecture Only"

steckel:
Defer to iOS default settings for "Build Active Architecture Only"
which allows multiple architectures to be built with the Debug build
configuration.

Many users of KIF are being forced to fork in order to enable building
for i386 architecture. By removing this restriction, there will be a
larger compatibility with existing applications looking to add
functional testing with KIF.


steckel:
FYI: Tests pass for me locally. Seems to just be flakey.

phatmann:
Tested this out, and everything seems to still build ok for me, so I will go ahead and merge this. If this causes any issue for anyone, please let me know.

steckel:
Thank you!

Issue #464, opened 2014-09-22
accessibilityElementMatchingBlock ends up in a loop


tapwork:
We have some `UICollectionView` tests. There we have a strange behavior inside `UIView (KIFAdditions)` at `- (UIAccessibilityElement *)accessibilityElementMatchingBlock:(BOOL(^)(UIAccessibilityElement *))matchBlock`.
We are waiting for a specific cell in the collectionView. The cell in the collectionView with the `UIAccessibilityLabel` is in the visible bounds, but KIF won't see / find it.
Obviously the method goes through all sections and items of the UICollectionView. KIF cannot find the cell and ends up in a loop with a huge backtrace. 

Did you encounter this issue before?

A fix for looping could be a timeout for that execution block. But this is not the real fix. Actually KIF should find the element. (If not KIF should fail after a timeout)
![screen shot 2014-09-22 at 17 43 01](https://cloud.githubusercontent.com/assets/397342/4359085/406c5ab6-426f-11e4-8ae4-f0692e9d93e3.png)

phatmann:
Fixed by #605

ninjazoete:
Hi, same happened to us today, on iphone 4s it was in infinite loop when using UIAccessibilityLabel. On iphone 5, 6, 6+ it was fine. We changed implementation to only base the navigation in this component on index paths. Now it is fine. I have this version with your fix and it still did not continue to next element so there must be a case where this bug is still occuring.

ninjazoete:
* thread #1: tid = 0x124e65, 0x053199ce libsystem_kernel.dylib`mach_msg_trap + 10, queue = 'com.apple.main-thread', stop reason = signal SIGSTOP
  * frame #0: 0x053199ce libsystem_kernel.dylib`mach_msg_trap + 10
    frame #1: 0x05318a70 libsystem_kernel.dylib`mach_msg + 68
    frame #2: 0x046dceb6 CoreFoundation`__CFRunLoopServiceMachPort + 214
    frame #3: 0x046dc298 CoreFoundation`__CFRunLoopRun + 1400
    frame #4: 0x046dba5b CoreFoundation`CFRunLoopRunSpecific + 443
    frame #5: 0x046db88b CoreFoundation`CFRunLoopRunInMode + 123
[...]

PR #605, opened 2015-03-20
Fixes issue #464 (infinite recursion in UICollectionView)

mfclarke:
In my case, the UICollectionView was being searched for a cell, but was scrolling to a cell who's frame size was 0 over and over again. This PR fixes infinite recursion in those cases.


mfclarke:
resolves #464

phatmann:
Excellent! So glad to see #464 fixed.

PR #606, opened 2015-03-23
Syntax highlight Objective-C in README

evanwalsh:
Hello! I noticed the Objective-C code in the README wasn't syntax highlighted, but the other code blocks were. This PR fixes that!


phatmann:
Nice catch, thanks.

Issue #590, opened 2015-03-10
Picker tests failing


phatmann:
-[PickerTests testSelectingDateTime] failed: The step timed out after 10.00 seconds: Found an accessibility element with the label "Date Time Selection", but with the value "Monday, Jun 17, 06:43 AM", not "Sunday, Jun 17, 06:43 AM"

PR #607, opened 2015-03-24
Added printViewHierarchy to the UIView category. 

SebastianMecklenburg:
It prints the view hierarchy along with KIF related information, like the accessibility label or identifier, whether the view is visible or not etc.. I found it quite useful for debugging when the waitForView... methods fail for unknown reasons. 
The methods comes in two variants, an instance method which prints the hierarchy starting with the given view and a class method, called with ```[UIView printViewHierarchy]``` which starts with the top window(s) and prints all the subviews from there.

### DIFF
--- a/Additions/UIView-KIFAdditions.h
+++ b/Additions/UIView-KIFAdditions.h
@@ -115,4 +115,14 @@ typedef CGPoint KIFDisplacement;
  */
 @property (nonatomic, readonly) UIWindow *windowOrIdentityWindow;
 
+/*!
+ @abstract Prints the view hiererchy, starting from the top window(s), along with accessibility information, which is more related to KIF than the usual information given by the 'description' method.
+ */
++(void)printViewHierarchy;
+
+/*!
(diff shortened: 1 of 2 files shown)


SebastianMecklenburg:
The CI fail is due to a failed test in the UIPicker test suite and is not related to my addition.

SebastianMecklenburg:
Oh, it can also be called on the fly from the debugger console with 

```
po [UIView printViewHierarchy]
```
or
```
po [someUIViewInstance printViewHierarchy]
```

phatmann:
Tests were failing due to #590. Now fixed.

phatmann:
@SebastianMecklenburg, this looks cool and useful, thanks for submitting it.

I need 2 changes before I can merge this:

1. There are 2 commits. Please squash the commits into 1.
2. Make the first line of the commit message 50 characters or less. A longer description is good, but the first line should be short. Github will display the first line as the commit "name", and then clicking the commit will show the whole commit message.

You can do both of these tasks via `git rebase -i`.

SebastianMecklenburg:
Righto, done.

phatmann:
Nice work, thanks!

PR #613, opened 2015-03-30
only include visible cells when matching accessibility elements

cleverandson:
The method which matches accessibility elements

```
- (UIAccessibilityElement *)accessibilityElementMatchingBlock:(BOOL(^)(UIAccessibilityElement *))matchBlock;
```

matches against all accessibility elements found in a table view which are not visible on the screen. This is probably a typo, so the negation was added to match only against elements which are visible on the screen.


phatmann:
It is not a typo. See the comment above your change: 

   // Skip visible items because they are already handled

Visible items are already accessible via the normal accessibility mechanism. This loop is just hunting for invisible ones.

cleverandson:
this decision renders the kif project unusable for us unfortunately. since we have a collection view which reloads automatically more items if all the loaded items are visited. so every time accessibilityElementMatchingBlock gets called it loads more cells. only after a small test we ended up with more then >1000 cells.

phatmann:
Are the extra cells being loaded because no visible cell matched a given accessibilityLabel?

justinseanmartin:
No updates since 2015, going to close this. Would be happy to revisit if the questions above are answered.

PR #614, opened 2015-03-31
introducing swipeAccessibilityElement to provide ability for complex logic

ohenak:
* making swipeViewWithAccessibilityLabel:_ reusing swipeAccessibilityElement
* following the same pattern as tap, wait and scroll.


phatmann:
Looks good, thanks.

PR #615, opened 2015-04-02
Fixed: UIPanGestureRecognizer velocity value

kdlogen:
This little fix is related to **UIPanGestureRecognizer**.
The main problem was that the _UIEvent_ object doesn\`t have a correct time stamp. That was always 0. If the time stamp is constant than the ```[panGestureRecognizer velocityInView:view]``` value remains always 0 on dragging the generated _UITouch_ objects from one point to another.

This fix synchronizes _UITouch_ and _UIEvent_ objects time stamp, and with this little change the system provides a valid velocity value when you call ```velocityInView:``` method.

Best regards,
LoGeN

### DIFF
--- a/Additions/UIView-KIFAdditions.m
+++ b/Additions/UIView-KIFAdditions.m
@@ -485,6 +485,8 @@ NS_INLINE BOOL StringsMatchExceptLineBreaks(NSString *expected, NSString *actual
             }
             UIEvent *eventDown = [self eventWithTouches:[NSArray arrayWithArray:touches]];
             [[UIApplication sharedApplication] sendEvent:eventDown];
+            
+            CFRunLoopRunInMode(UIApplicationCurrentRunMode, DRAG_TOUCH_DELAY, false);
         }
         else
         {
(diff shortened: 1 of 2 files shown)


Additions/UIView-KIFAdditions.m
```diff
@@ -485,6 +485,8 @@ - (void)dragPointsAlongPaths:(NSArray *)arrayOfPaths {
             }
             UIEvent *eventDown = [self eventWithTouches:[NSArray arrayWithArray:touches]];
             [[UIApplication sharedApplication] sendEvent:eventDown];
+            
+            CFRunLoopRunInMode(UIApplicationCurrentRunMode, DRAG_TOUCH_DELAY, false);
```
phatmann: In your comments for this pull request you talk about setting the event timestamp, which makes perfect sense. On the other hand you do not mention pumping the run loop. Can you explain what this line is for?

kdlogen: That little delay is needed because the "touch begin" and "touch moved" events  are to close to each other. If you would like to activate or deactivate the *UIPanGestureRecognizer* in method ```gestureRecognizerShouldBegin:``` depending on it`s velocity value, it can be 0 again.

Actually that modification isn`t tested, but it seams more logical.

phatmann:
Thanks for the explanation. Will merge now.

PR #619, opened 2015-04-05
Fixed static analysis error with uninitialized data

joeblau:
The current implementation for the gesture `timeStamp` uses the `mach_absolute_time()`.  This poses a problem because the Static Analyzer can't correctly cast `mach_absolute_time` to `AbsoluteTime` resulting in a Static Analyzer Issue.

screen shot 2015-04-05 at 10 29 50 am

This solution replaces `AbsoluteTime` with with `CFAbsoluteTime` and then uses the  `CACurrentMediaTime` [which is also based off of the internal clock](http://nshipster.com/benchmarking/) to set the timeStamp.

### DIFF
--- a/Additions/UIEvent+KIFAdditions.m
+++ b/Additions/UIEvent+KIFAdditions.m
@@ -127,7 +127,9 @@ typedef struct __GSEvent * GSEventRef;
 - (void)kif_setIOHIDEventWithTouches:(NSArray *)touches
 {
     uint64_t abTime = mach_absolute_time();
-    AbsoluteTime timeStamp = *(AbsoluteTime *) &abTime;
+    AbsoluteTime timeStamp;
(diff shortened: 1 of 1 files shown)


TBonnin:
Given that `CACurrentMediaTime` requires to import Quartzcore, I would prefer the following initialization rather than adding another framework dependency;
```
AbsoluteTime timeStamp;
timeStamp.hi = (UInt32)(abTime >> 32);
timeStamp.lo = (UInt32)(abTime);
```

https://gist.github.com/anonymous/50a24c00204423ee474e

joeblau:
Hehe, no problem, but I can't take the credit, you need to thank Xcode Server.  I ran my whole project through it and it exposed all of these issues with my app as well as every dependency.

I updated the PR.

TBonnin:
Thanks @joeblau 
@phatmann This change looks sane for merge ;)

PR #621, opened 2015-04-07
Remove non-visible table view cells so that they don't stick around.

hiltonc:
The naive approach to finding accessibility elements in a table view
results in cells being instantiated and added as subviews of the table
view, despite not being visible. When the table view has thousands of
rows with heavyweight cells this can cause severe memory pressure.
This commit uses an explicit approach to finding accessibility elements
in table views, which allows for proper removal of non-visible cells.


hiltonc:
Retriggering Travis CI.

hiltonc:
I keep iterating on this. I'll make a new PR when I'm satisfied with it.

PR #622, opened 2015-04-08
Remove non-visible table view cells so that they don't stick around.

hiltonc:
The implict approach to finding accessibility elements in a table view
results in cells being instantiated and added as subviews of the table
view, despite not being visible. When the table view has many rows
with heavyweight cells this can cause severe memory pressure. This
commit uses a direct approach to find accessibility elements in table
views, which allows for proper removal of non-visible cells. It also
has the benefit of returning actual views instead of accessibility
elements.

### DIFF
--- a/Additions/UIView-KIFAdditions.m
+++ b/Additions/UIView-KIFAdditions.m
@@ -121,7 +121,12 @@ NS_INLINE BOOL StringsMatchExceptLineBreaks(NSString *expected, NSString *actual
 
 - (UIAccessibilityElement *)accessibilityElementMatchingBlock:(BOOL(^)(UIAccessibilityElement *))matchBlock;
 {
-    if (self.hidden) {
+    return [self accessibilityElementMatchingBlock:matchBlock notHidden:YES];
+}
+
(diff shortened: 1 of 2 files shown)


phatmann:
Summing up, you are using the same technique for UITableView that we use with UICollectionView to locate cells. Looks good.

ksuther:
This PR broke searching for table view cells that aren't visible if the table cells are created programmatically. For example, if I have a cell with accessibility label "Table, Cell" and it isn't visible, it won't be found. The tests in the storyboard work because the cells created from the storyboard have accessibility elements as children. Programmatically created table view cells don't seem to have those same elements. I'm still investigating, but noting this here for now.

phatmann:
@ksuther Good to know. Any chance you can come up with a fix?

ksuther:
Unfortunately I don't have any time to look into this right now, I ended up just reverting to get things working again. If I get some free time to fix this I'll definitely send a PR, but I'm not expecting anything from myself soon.

phatmann:
Added #663 to cover @ksuther's issue.

PR #627, opened 2015-04-14
Move sys alert simulator macro to implementation

rymir:
So it doesn't affect the client code.

Otherwise in the tests we have to write

```
#if TARGET_IPHONE_SIMULATOR
tester().acknowledgeSystemAlert()
#endif
```

to make it build for the device.

### DIFF
--- a/Classes/KIFUITestActor.h
+++ b/Classes/KIFUITestActor.h
@@ -406,13 +406,11 @@ static inline KIFDisplacement KIFDisplacementForSwipingInDirection(KIFSwipeDirec
  */
 - (void)tapItemAtIndexPath:(NSIndexPath *)indexPath inCollectionViewWithAccessibilityIdentifier:(NSString *)identifier;
 
-#if TARGET_IPHONE_SIMULATOR
 /*!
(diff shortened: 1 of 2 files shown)


rymir:
By the way, in swift it looks like `#if TARGET_IPHONE_SIMULATOR` doesn't work, we could use `#if (arch(i386) || arch(x86_64)) && os(iOS)`

phatmann:
This is a good change, but let's switch to using `#if (arch(i386) || arch(x86_64)) && os(iOS)` as you suggest.

rymir:
When using `#if (arch(i386) || arch(x86_64)) && os(iOS)` in objc it complains that you cannot use arch() in preprocessor macros. Looks like we have to use it in the swift code, in objc we can rely on `#if TARGET_IPHONE_SIMULATOR` working. I would still move it from the interface to the implementation of course, what do you think?

phatmann:
See the comments in #520 on wrapping the relevant code in a `#if TARGET_IPHONE_SIMULATOR` block. What are your thoughts on the matter?

justinseanmartin:
No updates since 2015, going to close this.

PR #632, opened 2015-04-22
Fix memory leaks in UIEvent+KIFAdditions

TBonnin:


phatmann:
@TBonnin thanks for the quick fix!

PR #636, opened 2015-04-25
Merged master

simX:
A few improvements are included in this branch.  First and foremost: adds support for accessibility identifiers, so that proper accessibility support can co-exist with kif automation.  Second: adds a framework target so that kif can be built with carthage.  Other minor improvements as well.


phatmann:
Did you mean to submit this as a pull request? There is already support for identifiers in `KIFUITestActor-IdentifierTests` and a pull request for carthage is in the works.

simX:
@phatmann Sorry about the confusion.  I completely blanked on checking the current branch for accessibilityIdentifier support, and it looks like I started my branch for Carthage support before the other pull request was created.  So, yeah, this pull request is unnecessary.

PR #639, opened 2015-04-30
UIPanGestureRecognizer velocity test cases

kdlogen:
I implemented 4 new test cases for testing the UIPangestureRecognizer`s velocity value.

### DIFF
--- a/KIF Tests/GestureTests.m
+++ b/KIF Tests/GestureTests.m
@@ -9,6 +9,16 @@
 #import <KIF/KIF.h>
 #import <KIF/KIFTestStepValidation.h>
 #import <KIF/KIFUITestActor-IdentifierTests.h>
+#import <KIF/UIView-KIFAdditions.h>
+
+#define kPanMeAccessibilityString               @"Pan Me"
+#define kVelocityValueLabelAccessibilityString  @"velocityValueLabel"
+
+#define kPanLeftRegex                           @"^X:-[0-9\\.]+ Y:0.00$"
+#define kPanUpRegex                             @"^X:0.00 Y:-[0-9\\.]+$"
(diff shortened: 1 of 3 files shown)


phatmann:
I always like to have more tests, thanks!

kdlogen:
Your welcome! ;)

We wanted to mention that the tests didn't work if we requested the UILabel with method ```waitForViewWithAccessibilityLabel:``` (we used ```waitForTappableViewWithAccessibilityLabel:```). As we noticed this behaviour happens because somehow the gesture recognizer can't initialize everything before the test case starts. If you make an extra wait with ```CFRunLoopRunInMode``` (~0.1 sec!?) method call before sending every touch event that will work with ```waitForViewWithAccessibilityLabel:```(```UIView+KIFAdditions.h - dragPointsAlongPaths:``` method).

p.s.: You can try it by running only one older swipe test case. The test case should fail!

Does it have any sense? What do you think about it?

phatmann:
@kdlogen, I do not understand the issue you mention above. Why would the gesture recognizer prevent `waitForViewWithAccessibilityLabel:` from working? Did you mean to say that `waitForTappableViewWithAccessibilityLabel:` was failing?

kdlogen:
If you run a single test case (for example GestureTests.m -> testSwipingLeft) then that fails. You can try out waiting for view with method *[tester waitForViewWithAccessibilityLabel:@"Swipe Me"];* but that isn't enough, you should use *[tester waitForTappableViewWithAccessibilityLabel:@"Swipe Me"];* method.
I don't know exactly what is the main difference between those methods, but it seams the system can`t  initialize the gesture recognizers before swiping on the view.

You can find a solution in pull request #669.

phatmann:
@kdlogen Thanks for the fix.

PR #646, opened 2015-05-06
KIF v4

phatmann:
Contains all the proposed changes for KIF v4. All discussion about v4 should take place here. 

**This pull request will not be merged until all v4 work is complete.**


mokagio:
Thumbs up for this :+1:  I'll be using in in a couple of projects and hopefully I'll be able at least send some feedback.

phatmann:
Worth mentioning here: Xcode 7 includes UI testing as part of XCTest for iOS 9. This is fantastic news, and I would like to think that frameworks like KIF pushed Apple in the right direction. The most exciting part of this is that you can record UI actions and they are automatically turned into test code.

We need to decide the following:

* Is there still a place for KIF once the need for iOS 8 and below testing is diminished? 
* If so, do we rewrite the internals of KIF to use XCUIElement methods?

KIF could provide higher-level calls for folks that don't want to use recording. Not sure yet if there will be a large demand for this. We need to play with recording and see how the quality of the generated code is.

There is a nice write-up on UI testing on Xcode 7 [here](http://www.mokacoding.com/blog/xcode-7-ui-testing/).

bnickel:
Relevant: https://medium.com/@larcus94/ui-testing-with-xcode-7-221d16bad276

My initial thoughts are that the new tool largely replaces/improves on KIF but there are a few area where KIF wins. This will be similar to comparisons between KIF and existing frameworks based on the automation API. 

Where Apple wins:

- Tests are not hosted in the app and tests could recover better from failure. Similarly you're not abusing the run loop to make things work. 
- Tests use real automation API which should be faster/more reliable than KIF's searches. Better for collection/table views?
- Recording is okay and will get better. 

Where KIF wins:

- Access to the actual view hierarchy. Doesn't look like the the automation API can access that.
- Simulation of complex gestures and touch events. 
- Direct access to app objects (since running in app). NSNotification, openURL simulation, etc. Better mocking/control. 

Of course I may be wrong on some of these limitations. I've not proved too deeply.

phatmann:
@bnickel Thanks so much for your response. To sum up, the biggest differences between KIF and XUI are:

* The tests do not run in the app's process. No access to app instance, controller instances, etc.
* No access to actual view instances, just accessibility objects

Or summing up the summary: you have the same limitations in XUI that you have in UIAutomation. This is going to be a deal-breaker for many folks. So KIF is still relevant!

We should see if we can leverage XCUIElement's magic for our own evil purposes. I want to avoid tunneling between the two processes, because that is Subliminal's territory (for now).

justinseanmartin:
Completely agree with @phatmann. KIF is very much relevant with those limitations of XUI. For those exact reasons, we're sticking with functional tests written in KIF.

hiltonc:
A lot of the KIF testing I do takes advantage of accessing application data to identify which accessibility labels to tap. My apps are data heavy, and it is extremely useful to have a test step that says, "go to the screen for this content" without having to write out the 10 steps to get there (which would then be fragile as content is dynamic and gets moved around). XUI looks really great, but I don't have a tunneling solution so for now I'm sticking with KIF.

phatmann:
I was hoping perhaps that the Subliminal folks would be working on an XUI tunneling solution, but that project has not been updated in over a year. Maybe this is something to consider for KIF v4. That would be amazing if it could be pulled off.

phatmann:
I am closing this PR in favor in #794, as that contains the new `viewTester` that moves KIF toward v4.

PR #648, opened 2015-05-08
Move print view hierarchy methods

GraemeEArthur:
I ended up keeping the methods in a UIView category for better backward compatibility. I also split it out into slightly smaller methods for better readability.


phatmann:
Nice change, I am very glad to see this code pulled out into its own file.

However, before I merge it,  @GraemeEArthur can you squash the 8 commits down to one? Thanks.

GraemeEArthur:
That's it squashed and rebased. Seems there's a failing test but don't think it's related to this work.

phatmann:
Looks great, thanks again. These kind of changes keep KIF clean.

PR #651, opened 2015-05-13
Add the ability to background the app

jzucker2:
Uses UIAutomation's ability to background an app for a duration and bubbles that up to the dev for use in tests.


jzucker2:
Totally open to implementation detail changes, but put this up as a first pass. Overall it's pretty simple so shouldn't require many/if at all changes.

phatmann:
Cool to see the new UIAutomation integration being used for other useful features like this.

Like you, I am concerned about having a new Background controller just for this one test. Can you add the test to System Tests instead, with the notification handler in there?

jzucker2:
Ended up being slightly more complicated to test from that suite, but not too bad. Let me know what you think

jzucker2:
Also the CI is failing but not on my tests

phatmann:
Well, I can see that I steered you wrong on the tests :-( Having a separate view controller really was the cleanest solution. Any chance we can revert back to that?

KIF Tests/SystemTests.m
```diff
@@ -33,6 +33,14 @@ @interface SystemTests : SenTestCase
 
 @implementation SystemTests
 
+- (void)setUp {
+    [[NSNotificationCenter defaultCenter] postNotificationName:@"SystemTestsNotification" object:nil userInfo:@{@"setup" : @YES}];
+}
+
+- (void)tearDown {
+    [[NSNotificationCenter defaultCenter] postNotificationName:@"SystemTestsNotification" object:nil userInfo:@{@"setup" : @NO}];
```
phatmann: Did you mean to send "tearDown" here?

jzucker2: Ah, that was just being a little lazy when I tried to cram everything into the system tests controller. That code isn't necessary when I revert to the older solution.

jzucker2:
Had some merge conflicts, and decided to move to a different pull request https://github.com/kif-framework/KIF/pull/655

Issue #650, opened 2015-05-13
Carthage: Reason: The run destination iPad 2 is not valid for Running the scheme 'KIFFramework'.


safetyscissors:
Hi,

I recently tried to install KIF using Carthage and got the following error.

I'm currently using:

Xcode 6.3.1
Build version 6D1002

*** Fetching OHHTTPStubs
*** Fetching KIF
*** Fetching specta
*** Checking out KIF at "v3.2.2"
*** Checking out OHHTTPStubs at "4.0.1"
*** Checking out specta at "v1.0.0"
*** xcodebuild output can be found in /var/folders/rr/z_x95s291hq84njdww4flnq80000gp/T/carthage-xcodebuild.qAcfdL.log
*** Building scheme "KIFFramework" in KIF.xcodeproj
xcodebuild: error: Failed to build project KIF with scheme KIFFramework.
	Reason: The run destination iPad 2 is not valid for Running the scheme 'KIFFramework'.

Looking at the log I have this:

Build settings from command line:
    SDKROOT = iphonesimulator8.3

jberkel:
encountered the same problem, the issue might be that the test host target is included in the framework target (it's probably not needed there). for some reason xcode sometimes picks the first installed simulator which happens to be 'ipad 2'.

phatmann:
@jberkel Glad you have an idea here. Are you going to submit a pull request with your fix in it? Or do you need to test it more?

jberkel:
need to test a bit more but will submit PR once i can confirm it

simX:
I can confirm that deleting the "Test Host" reference in the KIFFramework scheme fixes this problem.

phatmann:
Fixed by #652

PR #652, opened 2015-05-14
Remove test host reference

jberkel:
Possible fix for #650


phatmann:
Thanks for researching this and coming up with a clean solution.

jberkel:
i keep getting emails from xcodeserver about failing builds – any way to switch this off?  
"Integration #46 of integration tests nightly finished with 1 error."

phatmann:
I get the same darn emails and I cannot figure out how to stop them. Clearly some issue with the Travis/Github integration.

bnickel:
Welcome to the club. Someone has set up set up an Xcode bot and set up their build to keep pulling KIF's `origin/master`. (Does Cocoapods `:head` do this?)  It'll eventually go away when the build gets fixed and you fall out of the committers list.

If you can see other committer addresses in the email, I recommend complaining to them about their email notification config.  If it's just ServerName.local then that just sucks.

PR #624, opened 2015-04-12
Carthage

lottadot:
Implements [KIF/KIF#552](https://github.com/kif-framework/KIF/issues/552).

To verify:

#1: Grab this KIF fork and run unit tests:

````
git clone https://github.com/lottadot/KIF.git
cd KIF
git checkout carthage
xcodebuild test -scheme KIFFramework -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 6,OS=latest'
````

#2: Grab example project:

````
brew install Carthage
git clone https://github.com/lottadot/KIF-Carthage-Example.git
cd KIF-Carthage-Example
carthage checkout
xcrun xcodebuild test -workspace KIF-Carthage-Example.xcworkspace -scheme ExampleAcceptanceTests -sdk iphonesimulator -destination 'platform=iOS Simulator,name=iPhone 6,OS=latest'
````

You’ll note that we aren’t doing `carthage bootstrap`. That’s intentional (for now). We cannot yet build KIFFramework via Carthage that is app-store Submittable (iOS8 Dynamic Frameworks must be code signed, and that is where Carthage has complications in trying to determine when to (not)sign, what to strip out of the binary, when, etc). 

I realize that most people do not submit to the App store w/ a build target that has KIF in it (hopefully). However, if we try to use the release version of Carthage and build we are shown code-siging issues (which Carthage has issues discussing too). See note in [KIF-Carthage-Example-README](https://github.com/lottadot/KIF-Carthage-Example/blob/master/README.md)

Here's the result of that, for the record:

````
git clone https://github.com/lottadot/KIF-Carthage-Example.git
cd KIF-Carthage-Example
$ carthage bootstrap --verbose
*** No Cartfile.resolved found, updating dependencies
*** Fetching KIF
*** Checking out KIF at "e66d724a01cc0b5911b5b583f894e478df8cb1e7"
*** Building scheme "KIFFramework" in KIF.xcodeproj
Build settings from command line:
    SDKROOT = iphonesimulator8.3
....
=== BUILD TARGET KIFFrameworkTests OF PROJECT KIF WITH CONFIGURATION Release ===

Check dependencies
`CodeSign error: code signing is required for product type 'Unit Test Bundle' in SDK 'iOS 8.3'`

2015-04-12 11:25:06.244 xcodebuild[17168:379933]  DeveloperPortal: Using pre-existing current store at URL (DeveloperPortal%206.3.db).
** BUILD FAILED **
````

PR #654, opened 2015-05-15
rename framework target to just 'KIF' so it can coexist with Cocoapods

simX:
We're exploring Carthage support in our app, and it gives us a lot of benefits now, but we can't drop iOS 7 support yet, which means we still have to support Cocoapods.

With the current Carthage support merged into KIF, you have to use the following syntax to import KIF header files: #import \<KIFFramework/KIF.h\>.  Unfortunately, if you use that syntax in source code that is also built via Cocoapods, the Cocoapods build fails.

In this pull request, I propose changing the name of the built product to simply "KIF.framework".  This is simpler and cleaner.  In addition, it allows you to use this syntax: #import \<KIF/KIF.h\>, which works for both Carthage and Cocoapods, regardless of which dependency manager you're using.  That way, you don't have to use any ugly macros to support both.

This pull request leaves the scheme and target names at "KIFFramework" (contrary to the somewhat inaccurate commit message), so that it's still understandable what that target and scheme are for.  It's just changing the product name to "KIF", which results in the built framework product being named "KIF.framework".

In addition, this pull request includes \<KIF/KIFUITestActor-IdentifierTests.h\> in the public header so that those of us that need to use accessibility identifiers can do so with the framework target.


simX:
cc-ing @lottadot and @phatmann since they worked on the original Carthage pull request: https://github.com/kif-framework/KIF/pull/624

phatmann:
Seems like a good change. I like the clean solution to a potentially thorny problem.

phatmann:
@simX, unfortunately this PR needs rebasing on master. Can you do that? If not I can do it.

simX:
Rebased, but had to create a new pull request: https://github.com/kif-framework/KIF/pull/657

PR #656, opened 2015-05-18
Use [UITableView visibleCells] instead of indexPathsForVisibleRows

TBonnin:
I bumped into a rare problem while trying to update to KIF 3.2.3

[UITableView indexPathsForVisibleRows] seems to return index paths for
non visible cells if screen folds on the tableview header.
I opened a radar with Apple. See http://openradar.appspot.com/radar?id=5191284490764288
To avoid this bug, we are now using the visibleCells method which
doesn't suffer from the same problem.

It won't be as fast as before since enumerating through the visible cells is necessary to get all the index paths but correctness is definitely the priority.
Let me know what you think. Thx


phatmann:
I agreed, best to avoid any weird situation. Might even fix some existing table-related bugs. Thanks!

PR #664, opened 2015-06-02
Removing semicolons from method signatures

kleinlieu:
The compiler does not complain about this, but it's nice to be consistent about it!

### DIFF
--- a/Classes/KIFUITestActor.m
+++ b/Classes/KIFUITestActor.m
@@ -320,13 +320,13 @@
     }];
 }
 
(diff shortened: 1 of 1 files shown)


mokagio:
It's the little things right? :smile: :+1:

kleinlieu:
@mokagio glad you like it :)

phatmann:
I figured it was a style thing, almost a protest: I *will* copy my function signature from the header file without removing the semi-colon! :-)

PR #668, opened 2015-06-08
Adding in new test methods for UIStepper

mikelupo:
adds in functionality for:
tapStepperWithAccessibilityLabel
tapStepperWithAccessibilityIdentifier

Adds new tapping test cases for the same.
To support the test cases new UI elements (A UIStepper and a UILabel)
were added to the storyboard.

### DIFF
--- a/Classes/KIFUITestActor.h
+++ b/Classes/KIFUITestActor.h
@@ -39,6 +39,17 @@ typedef NS_ENUM(NSUInteger, KIFPickerType) {
     KIFUIDatePicker
 };
 
+/*!
+ @enum KIFStepperDirection
+ @abstract Direction in which to increment or decrement the stepper.
+ @constant KIFStepperDirectionIncrement Increment the stepper
+ @constant KIFUIDatePicker Decrement the stepper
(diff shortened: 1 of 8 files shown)


phatmann:
I am glad to see new methods for stepper being added. 

In order to avoid duplication between the accessibility label and the accessibility identifier tests, there should be a method that operates on a view instance that contains the common code. Look at the other tests to see what I mean.

I am also concerned about assuming that the "-" button is on the left. Is there a way to actually find out which button decrements and which increments? I don't care if it is hacky, uses private methods, etc.

mikelupo:
@phatmann, 
I'll look at the de-duplication of code. 
To your comment about the "-" button being on the left. I don't think we can set it to another orientation. Am I wrong about this? Otherwise, I think we can just omit that text.

mikelupo:
Once again, I screwed up the tip of my branch. I'll close this pull request and resubmit from a new branch. Sorry about that. I'm very accustomed to the Gerrit workflow, however, the github workflow is still a bewilderment to me. :)

PR #673, opened 2015-06-10
add method swipeScreenAtPoint

anwol004:


phatmann:
@anwol004, I like this new addition to KIF. But I am wondering if we can factor out the common code between `tapScreenAtPoint` and `swipeScreenAtPoint`?

justinseanmartin:
No updates since 2015, going to close this. Would be happy to revisit if the question above is answered and a test is added to validate the functionality.

PR #675, opened 2015-06-11
add Xcode 7 compatibility

sync:
Xcode 7 wants
```
@property(strong) UIView *view;		
@property(strong) UIWindow *window;
```

Xcode 6 wants
```
@property(retain) UIView *view;		
@property(retain) UIWindow *window;
```

`strong` isn't valid in Xcode 6 so best is to just declare setters instead making it work with Xcode 6 and 7

### DIFF
--- a/Additions/UITouch-KIFAdditions.m
+++ b/Additions/UITouch-KIFAdditions.m
@@ -26,10 +26,10 @@ typedef struct {
 @property(assign) BOOL isTap;
 @property(assign) NSUInteger tapCount;
 @property(assign) UITouchPhase phase;
-@property(retain) UIView *view;
-@property(retain) UIWindow *window;
 @property(assign) NSTimeInterval timestamp;
 
+- (void)setWindow:(UIWindow *)window;
+- (void)setView:(UIView *)view;
 - (void)setGestureView:(UIView *)view;
 - (void)_setLocationInWindow:(CGPoint)location resetPrevious:(BOOL)resetPrevious;
 - (void)_setIsFirstTouchForView:(BOOL)firstTouchForView;


phatmann:
`@property(strong)` works fine in Xcode 6. It has been available since ARC was introduced.

sync:
sorry I meant it doesn't compile on Xcode 6 
`property attribute in class extension does not match the primary class [-Werror,-Wproperty-attribute-mismatch]`

because of: `@property(nonatomic,readonly,retain) UIWindow    *window;`

sync:
also  this is only the top of the iceberg, not only KIF doesn't compile but it's asserting in a couple of places once running under iOS 9

sync:
https://github.com/ksenks/iOS9-Runtime-Headers

shoerob:
Appears to be a memory exception. Question is.. is it UIEvent, or UITouch causing this?

elitecoder:
I don't see a stack trace when the crash happens. Any ideas how to go about it (or investigate)?

czechboy0:
For me it crashes in `-[KIFUITestActor scrollViewWithAccessibilityLabel:byFractionOfSizeHorizontal:vertical:]`. 100% reproducible. Is there an issue for this?

strangewiz:
Looks like UITouch now needs it's own _HIDEvent.  Is there a different init that should be used?

TBonnin:
@strangewiz was on the right track.
I pulled together a PR to add compatibility with Xcode 7 / iOS 9 https://github.com/kif-framework/KIF/pull/689

It requires more work but I am sharing it here to get everybody feedback and help. 
Cheers,

strangewiz:
@TBonnin Any ideas what's breaking alert views?

TBonnin:
@strangewiz No idea yet. Tapping happens at the right place but it looks like UIAlertController views don't react to it :/

TBonnin:
@strangewiz Ok I found out why. With iOS 9 it looks like the UIAlertControllerView only reacts to long press. See my change https://github.com/kif-framework/KIF/pull/689/files#diff-bfb168d5954b87f88ef57cb2ca64bd30L203

72lions:
Any ideas when Xcode 7 is going to be fully supported?

phatmann:
@72lions Smart folks are working on it, but I do not have an ETA yet. There are lots of new issues to work through.

Baconthorpe:
Hi, just wanted to check in on Xcode 7 compatibility. Any information regarding the timeframe of the fix or any workarounds would be greatly appreciated.

Keep up the good work!

phatmann:
This is fixed by #689

Baconthorpe:
Thanks!

PR #677, opened 2015-06-14
Add getter/setter for default animationWaitingTimeout

plu:
Hi,

I'd like to introduce a new getter and setter for the `animationWaitingTimeout` property. In our KIF test suite we do something like this in `application:didFinishLaunchingWithOptions:`:

```objc
#ifdef EBK_DISABLE_ANIMATIONS
    [UIView setAnimationsEnabled:NO];
#endif
```

This already decreased the execution time of our test suite dramatically. Without anymations enabled, we don't really need `animationWaitingTimeout = 0.5f`. If we could configure this from the outside, it would decrease the execution time even more.

Thanks,
plu


plu:
Thank you for merging it :+1: !

Issue #683, opened 2015-06-23
NSInvalidArgumentException hit in UIView-KIFAdditions.m


mikelupo:
failed: caught "NSInvalidArgumentException", "-[NSISRestrictedToNonNegativeVariable setTextDelegate:]: unrecognized selector sent to instance 0x17483cb60"
(
    0   CoreFoundation                      0x00000001856b42f4 <redacted> + 160
    1   libobjc.A.dylib                     0x0000000196cc00e4 objc_exception_throw + 60
    2   CoreFoundation                      0x00000001856bb3a4 <redacted> + 0
    3   CoreFoundation                      0x00000001856b8154 <redacted> + 928
    4   CoreFoundation                      0x00000001855baccc _CF_forwarding_prep_0 + 92
    5   UIKit                               0x0000000183aa93a0 <redacted> + 216
    6   UIKit                               0x0000000183aad3c8 <redacted> + 548
    7   UIKit                               0x0000000183aaeabc <redacted> + 156
    8   GW_AutomationTests          0x0000000100d4e7b0 -[UIView(KIFAdditions) accessibilityElementMatchingBlock:notHidden:] + 1772
    9   GW_AutomationTests          0x0000000100d4eba8 -[UIView(KIFAdditions) accessibilityElementMatchingBlock:notHidden:] + 2788
    10  GW_AutomationTests          0x0000000100d4e090 -[UIView(KIFAdditions) accessibilityElementMatchingBlock:] + 88
    11  GW_AutomationTests          0x0000000100d4e3b8 -[UIView(KIFAdditions) accessibilityElementMatchingBlock:notHidden:] + 756
    12  GW_AutomationTests          0x0000000100d4e090 -[UIView(KIFAdditions) accessibilityElementMatchingBlock:] + 88
    13  GW_AutomationTests          0x0000000100d4e3b8 -[UIView(KIFAdditions) accessibilityElementMatchingBlock:notHidden:] + 756
    14  GW_AutomationTests          0x0000000100d4e090 -[UIView(KIFAdditions) accessibilityElementMatchingBlock:] + 88
    15  GW_AutomationTests          0x0000000100d4e3b8 -[UIView(KIFAdditions) accessibilityElementMatchingBlock:notHidden:] + 756
    16  GW_AutomationTests          0x0000000100d4e090 -[UIView(KIFAdditions) accessibilityElementMatchingBlock:] + 88
[...]

PR #684, opened 2015-06-23
Possible fix for NSInvalidArgumentException

mikelupo:
Sometimes some stale UI leaks (think long test run) and you get
an invalid pointer.

refer to:
https://github.com/kif-framework/KIF/issues/683


mikelupo:
Tony, don't integrate this.

PR #689, opened 2015-07-06
Xcode 7 and iOS 9 support

TBonnin:
The main change consists in setting a now required internal IOHIDEvent to UITouch objects. https://github.com/kif-framework/KIF/commit/43163c0ee6cfa8c91b67ead4049ae296a56f0e86
This is of course very fragile and I believe the long term solution would be to add support for the new UI Testing framework provided by Apple and rely on XCUIElement API to interact with UI elements, and getting rid of those private API calls that will break every time Apple will make changes.

While this change allows most of the tests to run on iOS 9, some tests are still failing. I added a `//TODO` to the tests that are not passing on iOS 9 at the moment. See https://github.com/TBonnin/KIF/commit/6f8e4dc1fdcf4acc7ea1c97ba585edc076cd7671

~~Also I noticed I need to have the Accessibility Inspector enabled for KIF to work properly. I didn't look too much into it but maybe the AppSupport Library is not loaded properly.~~ See [below](https://github.com/kif-framework/KIF/pull/689#issuecomment-127792795)

TODO:
- ~~Fix failing tests~~ Wait for Apple to fix those issues:
   - ~~AccessibilityLabels are always nil on iOS 9 even when the `ApplicationAccessibilityEnabled` preference is enabled. I have opened a radar.~~ See [below](https://github.com/kif-framework/KIF/pull/689#issuecomment-127792795). It looks like we need to [enable the Accessibility Inspector](https://github.com/TBonnin/KIF/commit/e2da1dc19acb7e85bb3d2b63fcb0029dabcc7a99) on iOS 9
   - [testBackgroundApp](https://github.com/kif-framework/KIF/blob/aa97c5d74649c027401da8d7d39ed20504d74cbb/KIF%20Tests/BackgroundTests.m#L25) fails with the following error`caught "kUIAExceptionUnexpected", "-[UIAElementNil _prepareForAction:]: unrecognized selector sent to instance...` I tried to run a simple UIAutomation script with `target.deactivateAppForDuration(5);` and the same error is thrown. Probably a bug in iOS 9. I have filled a radar.
   - [testTappingLastRowAndSection](https://github.com/kif-framework/KIF/blob/aa97c5d74649c027401da8d7d39ed20504d74cbb/KIF%20Tests/TableViewTests.m#L26) and [testTappingRows](https://github.com/kif-framework/KIF/blob/aa97c5d74649c027401da8d7d39ed20504d74cbb/KIF%20Tests/TableViewTests.m#L28) fail. UITableViewCell.accessibilityTraits is incorrect on iOS 9 after tapping the cell (`UIAccessibilityTraitNone` instead of `UIAccessibilityTraitSelected`). I believe this is a bug in the iOS 9 beta. I have opened a radar.
- Exhaustive testing with iOS 8 and iOS 9

### DIFF
--- a/Classes/KIFUITestActor.m
+++ b/Classes/KIFUITestActor.m
@@ -204,6 +204,10 @@
         
         // This is mostly redundant of the test in _accessibilityElementWithLabel:
         KIFTestWaitCondition(!isnan(tappablePointInElement.x), error, @"View is not tappable");
+        
+        if ([NSStringFromClass([view class]) isEqualToString:@"_UIAlertControllerActionView"]) {
+            [view longPressAtPoint:tappablePointInElement duration:0.1];
+        }
         [view tapAtPoint:tappablePointInElement];
         
         KIFTestCondition(![view canBecomeFirstResponder] || [view isDescendantOfFirstResponder], error, @"Failed to make the view into the first responder");
--- a/Additions/UIApplication-KIFAdditions.h
+++ b/Additions/UIApplication-KIFAdditions.h
@@ -105,8 +105,6 @@ UIKIT_EXTERN NSString *const UIApplicationOpenedURLKey;
 @end
 
 @interface UIApplication (Private)
-- (BOOL)rotateIfNeeded:(UIDeviceOrientation)orientation;
-- (void)rotateIfNeeded:(UIDeviceOrientation)orientation completion:(void (^)(void))completion;
 - (UIWindow *)statusBarWindow;
 @property(getter=isStatusBarHidden) BOOL statusBarHidden;
 @end
--- a/Additions/UIEvent+KIFAdditions.m
+++ b/Additions/UIEvent+KIFAdditions.m
@@ -8,54 +8,9 @@
 
 #import "UIEvent+KIFAdditions.h"
 #import "LoadableCategory.h"
-#import <mach/mach_time.h>
+#import "IOHIDEvent+KIF.h"
 
 MAKE_CATEGORIES_LOADABLE(UIEvent_KIFAdditions)
-
-/* IOKit Private Headers */
-#ifdef __LP64__
-typedef double IOHIDFloat;
-#else
-typedef float IOHIDFloat;
-#endif
-typedef struct __IOHIDEvent * IOHIDEventRef;
-typedef UInt32 IOOptionBits;
-typedef uint32_t IOHIDDigitizerTransducerType;
-void IOHIDEventAppendEvent(IOHIDEventRef event, IOHIDEventRef childEvent);
-enum {
-    kIOHIDDigitizerTransducerTypeStylus  = 0x20,
-    kIOHIDDigitizerTransducerTypePuck,
-    kIOHIDDigitizerTransducerTypeFinger,
-    kIOHIDDigitizerTransducerTypeHand
-};
-enum {
-    kIOHIDDigitizerEventRange                               = 0x00000001,
-    kIOHIDDigitizerEventTouch                               = 0x00000002,
-    kIOHIDDigitizerEventPosition                            = 0x00000004,
-    kIOHIDDigitizerEventStop                                = 0x00000008,
-    kIOHIDDigitizerEventPeak                                = 0x00000010,
-    kIOHIDDigitizerEventIdentity                            = 0x00000020,
-    kIOHIDDigitizerEventAttribute                           = 0x00000040,
-    kIOHIDDigitizerEventCancel                              = 0x00000080,
-    kIOHIDDigitizerEventStart                               = 0x00000100,
-    kIOHIDDigitizerEventResting                             = 0x00000200,
-    kIOHIDDigitizerEventSwipeUp                             = 0x01000000,
-    kIOHIDDigitizerEventSwipeDown                           = 0x02000000,
-    kIOHIDDigitizerEventSwipeLeft                           = 0x04000000,
-    kIOHIDDigitizerEventSwipeRight                          = 0x08000000,
-    kIOHIDDigitizerEventSwipeMask                           = 0xFF000000,
-};
-IOHIDEventRef IOHIDEventCreateDigitizerEvent(CFAllocatorRef allocator, AbsoluteTime timeStamp, IOHIDDigitizerTransducerType type,
-                                             uint32_t index, uint32_t identity, uint32_t eventMask, uint32_t buttonMask,
-                                             IOHIDFloat x, IOHIDFloat y, IOHIDFloat z, IOHIDFloat tipPressure, IOHIDFloat barrelPressure,
-                                             Boolean range, Boolean touch, IOOptionBits options);
-IOHIDEventRef IOHIDEventCreateDigitizerFingerEventWithQuality(CFAllocatorRef allocator, AbsoluteTime timeStamp,
-                                                              uint32_t index, uint32_t identity, uint32_t eventMask,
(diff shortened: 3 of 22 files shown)


Classes/KIFUITestActor.m
```diff
@@ -204,6 +204,10 @@ - (void)tapAccessibilityElement:(UIAccessibilityElement *)element inView:(UIView
         
         // This is mostly redundant of the test in _accessibilityElementWithLabel:
         KIFTestWaitCondition(!isnan(tappablePointInElement.x), error, @"View is not tappable");
+        
+        if ([NSStringFromClass([view class]) isEqualToString:@"_UIAlertControllerActionView"]) {
+            [view longPressAtPoint:tappablePointInElement duration:0.1];
+        }
```
TBonnin: Checking the view class makes this code very fragile and prompt to fail if Apple changes the internal of UIAlertController
However we are already using this strategy in some other places. Ex: https://github.com/kif-framework/KIF/blob/master/Additions/UIView-KIFAdditions.m#L405

@phatmann and others let me know what you think. Thx

phatmann: Fine with me. Like you said, we do this all over the place.

TBonnin:
Update:
Just tested with Xcode 7 beta 4 and the first issue I mentioned in the description (ie: AccessibilityLabels are always nil on iOS 9) has been fixed. :)

The 2 other ones are still there.

TBonnin:
Update: 
Just tested Xcode 7 beta 4 and all 3 issues are still there. :(

strangewiz:
TBonnin@ can we either set AXInspectorEnabled (seems to require a simulator restart) or use the AXInspectorServices private framework [[[AXInspectorServices alloc] init] showInspector] ?

strangewiz:
Looks like I can get around the problem by calling [accessibilitySettingsPrefController setAXInspectorEnabled:@(YES) specifier:nil]; after loading AccessibilitySettings.bundle, added to _enableAccessibility

TBonnin:
Hi @strangewiz do you mind sending a patch or a gist my way ;)

strangewiz:
Here's a quick hack: https://gist.github.com/strangewiz/5bbcc0ff1934181d2afc

Pardon the lack of return checking, etc, as it is just a hack.

TBonnin:
Thanks @strangewiz I will check that when I have some time.

phatmann:
I am watching this conversation with interest. Let me know if I can help.

TBonnin:
I tested this quickly this afternoon. Thank you for the gist @strangewiz 
Enabling the Accessibility Inspector indeed allows KIF to access the accessibility labels. It is not as clean as the previous solution since the inspector opens automatically on top of the app but it could be a good fallback solution if Apple decide not to fix the pb with `ApplicationAccessibilityEnabled` (or have decided this is from now the new behavior)

TBonnin:
Response from Apple regarding the `iOS 9: Accessibility label of UI elements is always null` radar

_Apple Developer Relations 31-Jul-2015 02:17 PM_
> This issue behaves as intended based on the following:
> Accessibility only works when a feature using accessibility is enabled.

> We are now closing this bug report.
> If you have questions about the resolution, or if this is still a critical issue for you, then please update your bug report with that information.

> Please be sure to regularly check new Apple releases for any updates that might affect this issue.

_Apple Developer Relations 04-Aug-2015 12:55 PM_
> That means that asking for the accessibilityLabel of something like a button will not return a value unless VoiceOver or Switch control is running. 

> This is done as an optimization so that we don’t use and load a lot of accessibility code for the many devices that do not need it

> Please update your bug report to let us know if this is still an issue for you.

It looks like we must enable the Accessibility Inspector as proposed by @strangewiz

TBonnin:
The 2 other radars:
`UITableViewCell accessibilityTraits is incorrect when selected (iOS 9 only)`
`UIAutomation iOS 9 beta: UITarget.deactivateAppForDuration() fails with error`
are still open. I didn't have any response on them

phatmann:
@TBonnin thanks as always for staying on top of this. And kudos to Apple for timely responses to the radars.

TBonnin:
@phatmann I pushed the commit to enable the Accessibility Inspector. https://github.com/kif-framework/KIF/commit/aab1ffe7b062c382500ee8ffab025a799267b417 as proposed by @strangewiz 

However this solution is not going to work on real device (only simulator) since the Accessibility Inspector is not available on device.

Also I barely tested this PR on real devices. I would really appreciate if you could help on this matter.

strangewiz:
This problem only exists on the simulator -- device doesn't need any fix.

strangewiz:
Also, I wonder how the new UI testing works -- it doesn't seem to need to turn on the Accessibility Inspector.  I wonder if there's another way...

phatmann:
@TBonnin, are we ready to merge this PR yet? Specifically, after this PR is merged, will KIF build in Xcode 6 and 7, run in iOS 8, and more or less run in iOS 9? It does not have to be perfect in iOS 9, since we can file any outstanding issues separately.

I ask because there is a separate pull request (#700) to get KIF to build in Xcode 7. I would rather merge this PR since it includes that fix.

TBonnin:
@phatmann 
This branch builds in Xcode 6 and 7 (beta 5).
Based on my non exhaustive testing, 
- All but 3 tests are successful when run on iOS 9 simulator. I have added a comment to mark them in the code. https://github.com/TBonnin/KIF/commit/7074aa880656e923756bd6bddd5c8f889c28087e. I have downloaded and tested with Xcode 7 beta 5 today and those issues have not been fixed unfortunately. No update of the radar tickets neither. :(
- All tests pass when run on iOS 8 simulator. 
I have seen some flackiness on iOS 8 though with `testLongPressingViewWith...`, `testWaitingForSearchFieldToBecomeFirstResponder` and `testPickingAPhoto` when running the full suite. However those tests always pass when they or their respective test class are run individually. Maybe a state issue. What do you think?

I have not tested on a real device.

Also do you know why is Travis failing https://s3.amazonaws.com/archive.travis-ci.org/jobs/74443354/log.txt

Any help to run the KIF tests and cover as much device/simulator/version as possible would be appreciated. Thank you

phatmann:
@TBonnin Since this PR allows KIF to build in Xcode 6 & 7, we should merge it as soon as possible. Are you ready for me to do that? We can file the remaining iOS 9 issues as separate tickets.

* The failing tests in iOS 8: there has definitely been flakiness those three tests you mention so I don't think this is a new issue.
* Travis: it has been very flakey that last few months. Not sure whose fault it is. So I will say theirs :-)
* Real devices: we can capture any issues in separate tickets.

TBonnin:
@phatmann in this case this PR is ready for final review. Do you need me to squash all commits?

phatmann:
@TBonnin please arrange the commits as you see fit. There are so many different changes in different areas that having more than one commit for this PR is acceptable. 

And in case I forget to say it later: awesome work on this!

TBonnin:
Alright, all commits are self-contained. I won't squash them. Feel free to merge it.

phatmann:
Woot, KIF is ready for the future! A metric ton of thanks to @TBonnin. Someone needs to buy that man a beer :-)

TBonnin:
I have open 2 different tickets regarding the 2 remaining issues:
- `DeactivateAppForDuration:` is failing. https://github.com/kif-framework/KIF/issues/703
- Selected `UITableViewCell accessibilityTraits` is wrong. https://github.com/kif-framework/KIF/issues/704

phatmann:
Thanks @TBonnin!

PR #691, opened 2015-07-08
UIAutomationHelper.m: fix acknowledgeSystemAlert to close all alerts …

bagage:
…and not hang infinitely in case of multiple alerts

In my case, I have 2 notifications one above the other. The current script is looping indefinitely since `application.alert` is never nil after dismissing the first one.

I changed behavior so that all applications are closed; if you want to keep closing only the first alert, maybe you should consider changing return type to `bool` and returns `TRUE` if a popup was dismissed so that we can iterate until it returns `FALSE`?

### DIFF
--- a/Classes/KIFUITestActor.h
+++ b/Classes/KIFUITestActor.h
@@ -435,10 +435,10 @@ static inline KIFDisplacement KIFDisplacementForSwipingInDirection(KIFSwipeDirec
 
 #if TARGET_IPHONE_SIMULATOR
 /*!
- @abstract If present, dismisses a system alert with the last button, usually 'Allow'.
+ @abstract If present, dismisses a system alert with the last button, usually 'Allow'. Returns YES if a dialog was dismissed, NO otherwise.
  @discussion Use this to dissmiss a location services authorization dialog or a photos access dialog by tapping the 'Allow' button. No action is taken if no alert is present.
  */
-- (void)acknowledgeSystemAlert;
+- (BOOL)acknowledgeSystemAlert;
 #endif
 
 /*!
--- a/Classes/KIFUITestActor.m
+++ b/Classes/KIFUITestActor.m
@@ -774,8 +774,8 @@
     [self tapItemAtIndexPath:indexPath inCollectionView:collectionView];
 }
 
-- (void)acknowledgeSystemAlert {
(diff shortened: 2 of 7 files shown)


bagage:
OK this cannot work as is; for instance Gallery access alert will automatically open another alert after being dismissed and since `acknowledgeSystemAlert` do not check really alert class, we fail to dismiss the Gallery alert.
Still having the solution to return a boolean value or something else?

phatmann:
@bagage It seems like you do not want me to merge this yet?

bagage:
OK I modified function to return a BOOL instead. I merged 2 AlertViews tests into one, so that 2 alerts are displayed one above the other (so this is still checking these both tests, but in a single one).
I modified `ackgnoledgeSystemAlert` accordingly to properly detect that a popup has been dismissed. 

Does it look OK?

Classes/KIFUITestActor.h
```diff
@@ -76,7 +76,7 @@ static inline KIFDisplacement KIFDisplacementForSwipingInDirection(KIFSwipeDirec
 /*!
  @abstract Waits until a view or accessibility element is present.
  @discussion The view or accessibility element with the given label is found in the view hierarchy. If the element isn't found, then the step will attempt to wait until it is. Note that the view does not necessarily have to be visible on the screen, and may be behind another view or offscreen. Views with their hidden property set to YES are ignored.
-
```
phatmann: Your editor changed the whitespace of a bunch of lines. Please get rid of these changes so we get a clear diff for history.

phatmann:
Looks like a good change, clean and understandable.  A few things to do:

* I would prefer to have 3 unit tests: location authorization, notification scheduling and both. Then we can see what is failing.
* Get rid of all the whitespace changes
* Squash your commits

bagage:
* @phatmann : I merged 2 commits in once because permission is asked only once per application. I did not manage to find an easy way to "revoke" my own permission from within the application. Have you an idea for that?
* OK, will do that
* same.

bagage:
Done 2 & 3. 

Any news for point 1 @phatmann ?

phatmann:
@bagage, I am fine with the way you did the unit tests now that you explained your reasoning. Other changes look fine.

PR #692, opened 2015-07-08
Fix bad display for particular accessibilityLabel on error

### LINKED ISSUE #688: KIF waitForViewWithAccessibilityLabel does not match patterns having % correctly
Hi,

I am trying to wait for a label (using waitForViewWithAccessibilityLabel) having a pattern of the type -

SOMETHING, SOMETHING, New, New, ok, ok, 40%, 20%, 5°, 10° 

My NSLog shows the exact pattern when I print the variable. However, KIF throws the following error message -

Failed to find accessibility element with the label "SOMETHING, SOMETHING, New, New, ok, ok, 40, 20, 5°, 10°" 

Looks like the two '%' signs are not picked up for matching the pattern. This seems to be inconsistent behaviour because a couple of other tests work fine. I have checked the white spaces and string length many times for this issue but there is no difference.

Toldy:
I fixed #688 by adding an other failureInFile which not interpret the description.
For the test
```
waitForViewWithAccessibilityLabel("foobar%")
```
If you didn't define the accessibilityLabel "foobar%" in your project; failureInFile function was call with "foobar%" as formatString parameter hence the string was interpreted and the '%' deleted...

### DIFF
--- a/Additions/NSException-KIFAdditions.h
+++ b/Additions/NSException-KIFAdditions.h
@@ -10,6 +10,7 @@
 
 @interface NSException (KIFAdditions)
 
-+ (NSException *)failureInFile:(NSString *)file atLine:(NSInteger)line withDescription:(NSString *)formatString, ...;
++ (NSException *)failureInFile:(NSString *)file atLine:(NSInteger)line withDescriptionFormat:(NSString *)formatString, ...;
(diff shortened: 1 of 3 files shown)


phatmann:
Thanks for the fix. One request: please have `failureInFile:atLine:withDescriptionFormat:` call `failureInFile:atLine:withDescription:` instead of repeating the same code. (And please squash the commits or amend the current one).

Issue #598, opened 2015-03-16
No screenshot saved upon a failing test when using KIF with Specta


tutipeti:
Hey,

as Jenkins is giving me more and more headache I wanted to use your wonderful feature of saving screenshots for failed tests to know what state was the app then. Sadly this didn't work, no matter where was I setting the KIF_SCREENSHOTS variable.

I was able to pinpoint the issue to the fact that I'm using Specta here, as Specta is extending XCTestCase for the test files. Because of this no screenshot taking code is executed, as `- (void)failWithException:(NSException *)exception stopTest:(BOOL)stop;` is called in `XCTestCase-KIFAdditions` instead of `KIFTestCase`.

As far as I can tell @minhtule had the same issue in pull #436 - but that pull has been closed and I believe that the last comment has nothing to do with this issue.

Should I have missed something please tell me what else do I need to do to get screenshots taken while using Specta.

Thanks in advance,
Peter

phatmann:
If I understand the situation correctly, there are only two ways to fix this:

1. Change Specta so it does not break KIF
2. Call the screenshot code from `XCTestCase-KIFAdditions` and `SenTestCase-KIFAdditions`

Option 2 seems fine to me. Can anyone see any issues with it? It might result in a little code duplication, but I could live with this, since OCUnit is deprecated.

tutipeti:
If you are only awaiting my feedback here then I'm fine with Option 2 ;)

phatmann:
I will make the fix described in option 2

vase4kin:
Hello, guys! Any news?

phatmann:
No news yet. I have not made the fix described above. It should be simple but I have been too busy to work on it. If you want to submit a pull request with the fix, please go right ahead and I will merge it ASAP.

vase4kin:
@phatmann i created pull request #695. Could you please review?

PR #695, opened 2015-07-29
Added screenshot saving upon failure when using KIF with Specta

vase4kin:
Hello, guys!

This pull request is related to #598.


phatmann:
Thanks for the PR. But one change is needed: remove the `writeScreenshotForException` method and call from KIFTestCase. Otherwise non-Specta users will get two screenshots.

vase4kin:
Done, thanks for review.

PR #700, opened 2015-08-04
iOS 9 build compatibility.

dgatwood:
CHANGES:

* Modified the view and window property declarations to be compatible with the iOS 9 SDK.  The changes are protected by an ifdef so that it still builds against older SDKs.
* Xcode 7 throws errors if you have OCUnit targets even if you aren't using them, so I nuked the targets from orbit just to make sure.

### DIFF
--- a/Additions/UITouch-KIFAdditions.m
+++ b/Additions/UITouch-KIFAdditions.m
@@ -26,8 +26,14 @@ typedef struct {
 @property(assign) BOOL isTap;
 @property(assign) NSUInteger tapCount;
 @property(assign) UITouchPhase phase;
-@property(retain) UIView *view;
-@property(retain) UIWindow *window;
+
+#ifdef __IPHONE_9_0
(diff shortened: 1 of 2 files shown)


phatmann:
Even with your changes, KIF still will not work in iOS 9. See #689. Perhaps you want to join forces with that crew?

dgatwood:
We're currently planning to use Apple's UI testing for iOS 9 device testing.  However, we still need KIF for iOS 7/8 testing.  Right now, KIF breaks our build when building against the iOS 9 SDK, hence the patches.  Yes, it would be nice if we could use a single testing framework everywhere, but for now, being able to build without errors is enough.


David

phatmann:
PR #689 includes fixes to make KIF in Xcode 7. If that is going to be ready soon I will merge it and close this PR. If it looks like it will be a while then I will merge this PR or the relevant part of fa9da11e4 as a stopgap.

phatmann:
Fixed in #689

PR #711, opened 2015-08-28
README Missing Info for Swift: add missing import of KIF framework to use in swift projects

tiagomartinho:
without the import KIF statement I was not able to build the project, despite the import in the bridging header


phatmann:
Good catch, thanks for the addition to the README. Before I merge this, can you get rid of the whitespace change on line 103?

tiagomartinho:
done!

PR #713, opened 2015-08-31
Improve the tone when talking about accessibility

zamith:
I think the previous text implied a couple of things about accessibility that are not entirely true:

1. That accessibility only matters for sight impaired people
2. That accessibility is nothing more than adding labels

It is true that sight impaired people benefit from accessibility concerns, but so do people that are hard of hearing, with motor difficulties, dyslexia, etc. Also, as the linked documentation explains, there is more than adding labels, even that the most important thing **for KIF**, are the labels.

I'm open to suggestions on even better phrasings. :)


bnickel:
👍

phatmann:
I like this language better, thanks for the change.

PR #715, opened 2015-09-03
Actionsheet fix

plu:
I have added a test case first to prove its failure on iOS8 with the current state of the code. On iOS9 it was passing without any change, which was expected.

After that I have added a fix. I chose the same way of differentiating between iOS versions as I have found in other parts of this code base.


phatmann:
Great to see a clean fix with tests. Thanks!

PR #716, opened 2015-09-03
KIFUITestActor.m: fix enterText method to support unicode characters such as Emojis

bagage:
Previous method was not working properly when entering unicode characters such as Emojis; see screenshot 1 (old) vs 2 (new) to see the wrong rendering. I could not manage to reproduce it with a unit test in KIF project (but it is on linphone-iphone...), but at least it renders correctly now.
old old vs new new


phatmann:
I like this change, nice clean fix, thanks for doing it.

PR #717, opened 2015-09-07
Fix waiting for a nested UIScrollView

lbrndnr:
I had a problem where a `UICollectionViewCell` contained a `UICollectionView`. When I waited for the inner `UICollectionView`, the tester would scroll the outer `UICollectionView` horizontally to make the inner one visible. This did not work, as the inner view was already visible and not a direct child of the outer scroll view. To be precise, the tester scrolled the cell with the wanted `UICollectionView` out of bounds which broke the following assertions.

My fix is just one approach so I'm not sure how smart of an idea it is to solve it this way.

Cheers

### DIFF
--- a/Additions/UIAccessibilityElement-KIFAdditions.m
+++ b/Additions/UIAccessibilityElement-KIFAdditions.m
@@ -129,7 +129,7 @@ MAKE_CATEGORIES_LOADABLE(UIAccessibilityElement_KIFAdditions)
         if ([superview isKindOfClass:[UIScrollView class]]) {
             UIScrollView *scrollView = (UIScrollView *)superview;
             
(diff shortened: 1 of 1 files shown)


phatmann:
Thanks for the pull request.

I would rather have a specific check for the scenario in question. That is, we should specifically check if we have a UICollectionView inside a UICollectionViewCell instead of the more general (and confusing) `[view.superview isEqual:superview]`. We don't want to catch too many flies with this net :-)

justinseanmartin:
No updates since 2015, going to close this. Would be happy to revisit if the above comments were taken into account and a test was added to show the problem scenario and ensure that it doesn't regress in the future.

PR #722, opened 2015-09-22
Add new pullToRefresh methods with associated label and identifier tests

mikelupo:
New Pull To refresh convenience methods and associated test cases that can also serve as an example of how the methods can be used.

### DIFF
--- a/Classes/KIFUITestActor.h
+++ b/Classes/KIFUITestActor.h
@@ -50,6 +50,22 @@ typedef NS_ENUM(NSUInteger, KIFStepperDirection) {
 	KIFStepperDirectionDecrement
 };
 
+/*!
+ @enum KIFPullToRefreshTiming
+ @discussion The approximate time in which the pull to refresh travels downward.
+ @abstract Timing in which to perform the pull down before release.
+ @constant KIFPullToRefreshInAboutAHalfSecond about half a second
+ @constant KIFPullToRefreshInAboutOneSecond about one second
+ @constant KIFPullToRefreshInAboutTwoSeconds about two seconds
+ @constant KIFPullToRefreshInAboutThreeSeconds about three seconds.
+ */
+typedef NS_ENUM(NSUInteger, KIFPullToRefreshTiming) {
+	KIFPullToRefreshInAboutAHalfSecond = 20, //faster
+	KIFPullToRefreshInAboutOneSecond = 100,
+	KIFPullToRefreshInAboutTwoSeconds = 150,
+	KIFPullToRefreshInAboutThreeSeconds = 200, //slower
+};
+
 #define kKIFMajorSwipeDisplacement 200
 #define kKIFMinorSwipeDisplacement 5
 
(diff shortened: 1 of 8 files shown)


mikelupo:
@phatmann any chance you can review this PR? I'd like to get it in and take it into general use.
Thanks in advance

Classes/KIFUITestActor.h
```diff
@@ -50,6 +50,22 @@ typedef NS_ENUM(NSUInteger, KIFStepperDirection) {
...
+/*!
+ @enum KIFStepperPullToRefreshTiming
+ @discussion The approximate time in which the pull to refresh travels downward.
+ @abstract Timing in which to perform the pull down before release.
+ @constant KIFPullToRefreshInAboutAHalfSecond about half a second
+ @constant KIFPullToRefreshInAboutOneSecond about one second
+ @constant KIFPullToRefreshInAboutTwoSeconds about two seconds
+ @constant KIFPullToRefreshInAboutThreeSeconds about three seconds.
+ */
+typedef NS_ENUM(NSUInteger, KIFStepperPullToRefreshTiming) {
```
phatmann: I doubt you meant to call this `KIFStepperPullToRefreshTiming`. Looks like a copy/paste error from the enum above :-)

Classes/KIFUITestActor.h
```diff
@@ -477,6 +493,48 @@ static inline KIFDisplacement KIFDisplacementForSwipingInDirection(KIFSwipeDirec
...
+ */
+- (void)pullToRefreshViewWithAccessibilityLabel:(NSString *)label pullDownDuration:(KIFStepperPullToRefreshTiming) pullDownDuration;
+
+/*!
+ @abstract Pulls down on the view that enables the pull to refresh.
+ @discussion This will enact the pull to refresh by pulling down the distance of 1/2 the height of the view found by the accessibility label.
+ @param label The accessibility label of the view to swipe.
+ @param value The accessibility value of the view to swipe.
+ */
+- (void)pullToRefreshViewWithAccessibilityLabel:(NSString *)label value:(NSString *)value;
```
phatmann: Do we really need all three of the methods below in the public header file? I know that for methods like the "swipe" family we have done so, but for a high-level feature like pull to refresh I would think the two methods above would suffice.

Classes/KIFUITestActor.h
```diff
@@ -477,6 +493,48 @@ static inline KIFDisplacement KIFDisplacementForSwipingInDirection(KIFSwipeDirec
...
+ */
+- (void)pullToRefreshViewWithAccessibilityLabel:(NSString *)label;
+
+/*!
+ @abstract Pulls down on the view that enables the pull to refresh.
+ @discussion This will enact the pull to refresh by pulling down the distance of 1/2 the height of the view found by the accessibility label.
+ @param label The accessibility label of the view to swipe.
+ @param pullDownDuration The enum describing the approximate time for the pull down to travel the entire distance
+ */
+- (void)pullToRefreshViewWithAccessibilityLabel:(NSString *)label pullDownDuration:(KIFStepperPullToRefreshTiming) pullDownDuration;
```
phatmann: Why do we need to expose pull down duration? Do some views require more or less time?

Classes/KIFUITestActor.h
mikelupo: Thanks for your keen eye!



KIFStepperPullToRefreshTiming





-----Original Message-----

phatmann:
@mikelupo, thanks as always for another good change. I reviewed the code and added a few line comments.

Classes/KIFUITestActor.h
mikelupo: Re: Exposing the pulldown duration.


My thought was that the text string while the pull-down is in motion could be tested. If it's too quick, KIF will likely miss testing for the string in a [tester waitFor...]. 
Hence, this gives the test case author the ability to slow it down enough for KIF to perform the check.


In hindsight, is performing a check like that even possible while KIF is performing the pull-down?
Asked more succinctly, in your opinion, what's the most reliable way to implement the ability to test that?



Thanks,
Mike


-----Original Message-----

Classes/KIFUITestActor.h
mikelupo: I am able to remove 
- (void)pullToRefreshViewWithAccessibilityLabel:(NSString *)label value:(NSString *)value pullDownDuration:(KIFPullToRefreshTiming)pullDownDuration traits:(UIAccessibilityTraits)traits;


but below is needed for the identifier tests.

- (void)pullToRefreshAccessibilityElement:(UIAccessibilityElement *)element inView:(UIView *)viewToSwipe pullDownDuration:(KIFPullToRefreshTiming) pullDownDuration;






-----Original Message-----

mikelupo:
I think my response to the exposure of the pull down duration got lost in the mix. Curious if you missed that message. My question is in regards to the validity of doing so. 
So that KIF could test the string that is visible before the snap. But can KIF test the string while the pull down is still in motion? and if so...how?
If not, then I will remove that exposure if you really think it should not be there, IMHO, there is no harm in having the ability of slowing things down. The slowest setting is jittery but it works.
waiting on your response before I move forward with my changes.

Thanks in advance!

phatmann:
@mikelupo I don't think exposing the pulldown speed has much value. In order to test the text, as you propose, you would have to have KIF pull down halfway (for example), test the text, and then keep pulling down the rest of the way. If you want to expose that somehow, that would be an interesting addition. But definitely not a requirement.

mikelupo:
@phatmann I need some advice.
I am trying to add the functionality to check the pull down text while the control is in motion in similar fashion how we check the text while typing into a text field. This however is proving to be pretty challenging as - (void)dragPointsAlongPaths:(NSArray *)arrayOfPaths  (or a new method of similar flavor) would have to be able to check for the text. I think this is the right place to look as it's the only place KIF is aware that the looping drag motion in progress.
The method call I wanted to add inside dragPointsAlongPaths is: - (void)expectView:(UIView *)view toContainText:(NSString *)expectedResult.
What's the best way to make this happen?

phatmann:
@mikelupo Perhaps we can add a predicate callback function to dragPointsAlongPaths?

mikelupo:
Any specific examples of where this is being done?
The reason I ask is because if I add the block to the lower level code, I have to bubble the same up through the whole call stack. Which means that I have either overload the method calls, or add to the existing method calls and then add the parameter to all the other callers. Kind of ugly.


Unless there is a shortcut way of doing that? The code gets muddy fast with all those predicated methods.
I could use a notification perhaps?



-----Original Message-----

phatmann:
A notification is a much better idea than a callback! Good call.

mikelupo:
@phatmann I have updated this PR. Please have a look.

phatmann:
Looks good, thanks! Will merge now.

PR #723, opened 2015-09-30
Workaround broken `deactivateAppForDuration:` in iOS 9

### LINKED ISSUE #703: iOS 9: deactivateAppForDuration: fails
When executing `[tester deactivateAppForDuration:N]` the app is properly closed but doesn't reopen.

More info:
`deactivateAppForDuration:N` fails with the following error `kUIAExceptionUnexpected", "-[UIAElementNil _prepareForAction:]: unrecognized selector sent to instance`

I tried to run a simple UIAutomation script with target.deactivateAppForDuration(5); and the same error is thrown. Probably a bug in iOS 9. I have filled a radar, no response from Apple so far.

0xced:
-[UIATarget reactivateApp] was not rewritten for the new iOS 9 app switcher so we use a horrible hack to get the correct accessibility element.

There also seems to be an issue with the UIAApplication object returned by the `frontMostApp` method which has a nil name.

These hacks heavily rely on implementation detail of -[UIATarget deactivateApp] and -[UIATarget reactivateApp] and might break anytime.

Fixes #703

### DIFF
--- a/Classes/UIAutomationHelper.m
+++ b/Classes/UIAutomationHelper.m
@@ -7,11 +7,21 @@
 //
 
 #import "UIAutomationHelper.h"
-#include <dlfcn.h>
+#import <dlfcn.h>
+#import <objc/runtime.h>
 #import <UIView-KIFAdditions.h>
 
 @interface UIAElement : NSObject <NSCopying>
 - (void)tap;
+- (NSNumber *)pid;
+@end
+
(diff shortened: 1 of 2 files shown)


phatmann:
Wow! Great sleuthing. A very worthy hack indeed. Thanks for a fix that seemed impossible :-)

PR #727, opened 2015-10-14
Return the Accessbility Inspector to its previous state after tests

misterfifths:
This moves the code in KIFTestActor that enables the AX Inspector out to its own class that
is an XCTestObserver. When all tests finish, the AX Inspector is returned to its previous
state, saving a trip to the Settings app after every test.


phatmann:
Nice change. Clean and effective. Merging now.

PR #728, opened 2015-10-15
Restore iOS < 9 compatibility

0xced:
PR #727 uses feature only available with the iOS 9 SDK. This restores iOS 8 and earlier compatibility.


phatmann:
Great catch! Merging now.

PR #731, opened 2015-10-22
Correct README for test target creation in Xcode 7

obuseme:


obuseme:
This is literally just a README change, not sure why the build would have failed...

Issue #734, opened 2015-11-02
KIF and bitcode error


anas10:
XCTest framework doesn't have bitcode enabled. (Since Xcode 7 beta 3 it's treated as an error.)

![image](https://cloud.githubusercontent.com/assets/5764537/10882672/f99e5790-816c-11e5-8d33-4e960df5efb2.png)

Xcode 7.1
OSX 10.11.1
KIF (master 7e8df0980a39372424144730c899d94c3c167b1d)

anas10:
I just found out that this error seems to happen only when testing on a real device.

phatmann:
Closing this since we have #735 in play.

PR #735, opened 2015-11-02
Bitcode error on Xcode 7 and OSX 10.11

anas10:
Possible fix for issue #734


beatrichartz:
+1 KIF does not build with Carthage unless Bitcode is disabled

mikelupo:
@anas10 ,
I made a fix to the .travis.yml build script that will enable your change to compile. Can you rebase your change against upstream/master and then re-push? This way the parent to your change contains the fix.

mikelupo:
@phatmann this is still failing to build. Is it because of the code contained in this change or is it just travis being travis?

phatmann:
I have seen Travis fail when making README changes :-(

If Travis passes that is nice and I can merge, but when it fails I have to decide how to handle it. For small changes where the submitter has run the tests locally on their machine or otherwise verified it, I will go ahead and do the merge. For larger changes or changes that concern me, I will run the tests locally.

For this change I would merge it now, since it is small. was verified by the submitter...and is actually important to keep KIF working! :-)

mikelupo:
@phatmann Thanks. Merging!

anas10:
Didn't had time to rebase, you were to fast for me :) For changes like this it's ok though.  Thanks!

PR #737, opened 2015-11-07
Add method for waiting for the absence of a view identified through a predicate

Johennes:
This commit adds a few convenience methods for cases in which an element only has an accessibility hint but no label or identifier. I'd like to get some early feedback on whether or not something like this would be appreciated. I think for clients it's more handy to use these methods instead of fiddling with `NSPredicate`.

The commit also adds a generic `waitForAbsenceOfViewWithElementMatchingPredicate` method which did not exist before.

### DIFF
--- a/Classes/KIFUITestActor.h
+++ b/Classes/KIFUITestActor.h
@@ -205,6 +205,13 @@ static inline KIFDisplacement KIFDisplacementForSwipingInDirection(KIFSwipeDirec
 - (void)waitForAccessibilityElement:(UIAccessibilityElement **)element view:(out UIView **)view withElementMatchingPredicate:(NSPredicate *)predicate tappable:(BOOL)mustBeTappable;
 
 /*!
+ @abstract Waits until an accessibility element is no longer present.
+ @discussion The accessibility element matching the given predicate is found in the view hierarchy. If the element is found, then the step will attempt to wait until it isn't. Note that the associated view does not necessarily have to be visible on the screen, and may be behind another view or offscreen. Views with their hidden property set to YES are considered absent.
+ @param predicate The predicate to match.
+ */
+- (void)waitForAbsenceOfViewWithElementMatchingPredicate:(NSPredicate *)predicate;
+
+/*!
  @abstract Tries to guess if there are any unfinished animations and waits for a certain amount of time to let them finish.
  */
 - (void)waitForAnimationsToFinish;
--- a/Classes/KIFUITestActor.m
+++ b/Classes/KIFUITestActor.m
@@ -126,6 +126,29 @@
     }];
 }
 
+- (void)waitForAbsenceOfViewWithElementMatchingPredicate:(NSPredicate *)predicate {
+    [self runBlock:^KIFTestStepResult(NSError **error) {
+        // If the app is ignoring interaction events, then wait before doing our analysis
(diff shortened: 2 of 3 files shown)


bnickel:
Why would an element have a hint but not a label?

Johennes:
Good question! Sorry, I should have provided a little more background.

I ran into such a case when working on a UI test for Firefox for iOS ([see here](https://github.com/mozilla/firefox-ios/pull/1230)). The test tries to ensure the presence (or absence) of the blue buttons in the screenshot below.

simulator screenshot

These are search suggestions for the text entered into the URL bar. The buttons don't have an accessibility identifier (because their number may vary) and also don't possess an accessibility label (because their text is dynamic). All they got is the accessibility hint `"Searches for the suggestion"` which is used for searching the buttons in the test.

I couldn't find any better way for targeting these buttons and in the test I set up an extension on `KIFUITestActor` for waiting on a view with a hint.

```
private extension KIFUITestActor {
    func waitForViewWithAccessibilityHint(hint: String) -> UIView? {
        var view: UIView? = nil
        autoreleasepool {
            waitForAccessibilityElement(nil, view: &view, withElementMatchingPredicate: NSPredicate(format: "accessibilityHint = %@", hint), tappable: false)
        }
        return view
    }
}
```

It's not a big deal but I thought this might be useful to have in KIF itself.

bnickel:
Ah, so the view does have a label, you just have no guarantees as to what it may be.

I think the addition of `waitForAbsenceOfViewWithElementMatchingPredicate:` is a huge win.  I'm waffling on how broad the use case for an accessibility hint test is, though.  One concern is that we only cover a small segment of functionality compared to LabelTraitHint and Identifier which cover toggles, swipes, etc.  Deferring to @phatmann and @mikelupo.

justinseanmartin:
Changes to KIF seem reasonable. Hopefully the necessity for all of these methods will go away with the v4.0 API:

https://github.com/kif-framework/KIF/blob/v4/Classes/KIFUIViewTestActor.h

It looks like a method for accessibilityHint wasn't added yet in the v4 API, but it would be trivial to add.

It sounds like you don't have control over the accessibility of the items in search extensions, is that right? Do you have control over which items are returned in the search extension, for example by network mocking? An alternative would be spying the network request, and then searching for the text that you know should appear.

Johennes:
@bnickel: Yes, that's a valid concern. The use case might be too small to justify the maintenance overhead. I would be fine with striping the scope down to just `waitForAbsenceOfViewWithElementMatchingPredicate`.

@justinseanmartin: Actually we do have control about the accessibility of the blue suggestion buttons. However, currently we only set the accessibility hint. The UI test I was working on was an integration test that retrieved the suggestion items from the network. I guess technically it would be possible to mock the the suggestions. Selecting the views based on their hint did seem easier though.

PS: The v4.0 API looks very interesting! :)

phatmann:
@Johennes, do you plan on updating this PR per your comments above? An alternative is to submit a new PR once we have #794 merged with all the hint goodies you wanted added to `viewTester`.

Johennes:
@phatmann Yes, sorry for the lag. Based on the discussion above and in light of the upcoming v4 API, I stripped this PR down to just `waitForAbsenceOfViewWithElementMatchingPredicate` since that's the only functional addition.

phatmann:
Excellent, merging now!

PR #748, opened 2015-12-10
3D Touch Support (Work in Progress)

cratter49:
I created a new method which passes a force value to be initialized with a UITouch for triggering a 3d touch on a specified view. This force value is used in the creation of an IOHIDEvent which is also attached to a UITouch. This should theoretically trigger a 3D touch given the addition of these new pressure/force properties in the iOS 9+ runtime headers as well as my research of the structure of the IOHIDEvent struct. However, the end result of calling this method with any float value is that it synthesizes a touch that does not seem to trigger 3D touch. If anyone could help me figure out how to get this working I would greatly appreciate it. Thanks!

### DIFF
--- a/Additions/UITableView-KIFAdditions.m
+++ b/Additions/UITableView-KIFAdditions.m
@@ -43,7 +43,7 @@
     CGPoint destinationPoint = CGPointMake(sourcePoint.x, CGPointCenteredInRect([self rectForRowAtIndexPath:indexPath]).y);
     
     // Create the touch (there should only be one touch object for the whole drag)
-    UITouch *touch = [[UITouch alloc] initAtPoint:sourcePoint inView:self];
+    UITouch *touch = [[UITouch alloc] initAtPoint:sourcePoint inView:self withForce:0.0f];
     [touch setPhaseAndUpdateTimestamp:UITouchPhaseBegan];
     
     UIEvent *eventDown = [self eventWithTouch:touch];
--- a/Additions/UITouch-KIFAdditions.h
+++ b/Additions/UITouch-KIFAdditions.h
@@ -13,7 +13,7 @@
 @interface UITouch (KIFAdditions)
 
 - (id)initInView:(UIView *)view;
-- (id)initAtPoint:(CGPoint)point inView:(UIView *)view;
+- (id)initAtPoint:(CGPoint)point inView:(UIView *)view withForce:(float)force;
 
 - (void)setLocationInWindow:(CGPoint)location;
 - (void)setPhaseAndUpdateTimestamp:(UITouchPhase)phase;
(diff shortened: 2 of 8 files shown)


phatmann:
If we cannot get force touch working, an alternative is to simulate peek and pop. See these links for more info:

https://gist.github.com/nickfrey/07e2c6d8d2e5444fb91d
https://github.com/DeskConnect/SBShortcutMenuSimulator

cratter49:
Those examples are probably fine for simulating the peek and pop features of 3D touch in addition to shortcut menus but then we wouldn't have covered the case in which an app's view changes according to the amount of force applied to it. (Ex: The app's background/display changes color based off of the amount of force applied to a certain view.) I believe the method I previously attempted was on the right track and very close to triggering a 3D touch with varying force. If anyone could take a look at my current changes to KIF and give me any insight on whether or not my proposed solution would work based off of  knowledge of the UITouch and UIEvent objects or IOHIDEvents that would be great.

phatmann:
@TBonnin you are the resident expert of all things arcane and undocumented :-)

Any ideas about how to get force touch working?

TBonnin:
I will try to have a look before the end of the week.
@cratter49 Do you mind sharing the Tests for your changes? Thank you

TBonnin:
I had a quick look and to be honest I don't know how to add force touch support.
What I have noticed though is `pressure` and `force` are 2 different properties. (You can `NSLog(@"%lf %lf", touch.force, touch._hidEvent) if you are not convinced.`
Also `force` is not a value between `0` and `1.0` but between `0` and `maximumPossibleForce`.

cratter49:
Thanks for checking it out @TBonnin! 

I've been passing a variable amount of pressure to the IOHIDEvent object that's created and verifying that pressure is set in the created IOHIDEvent object but attaching it to a UITouch object and sending that to UIApplication doesn't seem to trigger a 3D Touch. 

My next two attempts will be to write to the read-only Force property of UITouch in memory by utilizing Objective-C's references and if that doesn't work then triggering the UIViewControllerPreviewingDelegate.

cratter49:
I made a breakthrough recently but was not able to get full 3D touch support. When attempting to set the pressure of a UITouch object I used KVC to bypass the need for a setter when writing to a read-only property as so: UITouch touch; [touch setValue:<pressure_value> forKey:@"pressure"]; When printing out the properties of my synthesized UITouch object the force value was set which triggers a basic 3D touch when attached to a UIEvent that is sent to the UIApplication. I've been using Apple's ViewControllerPreview iOS application for testing which demonstrates the Peek and Pop features of 3D touch. I currently can only trigger the Peeking action but not Pop. Once again, any further help on this would be greatly appreciated.

phatmann:
@cratter49 Glad to see progress. Do you want to try and merge what you have so far or wait until we have full functionality?

justinseanmartin:
Going to close this PR, as work on this seems to have been abandoned. Would be happy to review if someone wants to pick back up working on this.

PR #749, opened 2015-12-10
Optional line number on screenshot file name

Lucien:
The line number appended to the saved screenshot file name should be optional. Specify `0` or negative values to hide the appendage "`, line %lu`".

Added unit test.


phatmann:
Nice change, with tests! And the code is even cleaner now. Will merge.

PR #753, opened 2015-12-16
Update README

mikelupo:
Tony can you peruse this and merge if it's correct. 

Thanks,
Mike


phatmann:
Good catch! No reason to mention OCUnit in the README. People get that it is dead and gone.

PR #759, opened 2016-01-04
Fixed passing nil as pullDownDuration

benasher44:
This was causing a compiler warning about sending a void * to parameter of type KIFPullToRefreshTiming


phatmann:
Good catch. Will merge.

PR #760, opened 2016-01-05
Fixed IOHIDEvent memory leak

benasher44:
I also added the CF_RETURNS_RETAINED annotation, and then the static analyzer called out all of the leaks.


phatmann:
Nice catch! Will merge.

PR #763, opened 2016-01-06
Send 0 instead of nil as pullDownDuration fixing compilation error

ratkins:
Similar to https://github.com/kif-framework/KIF/pull/759


mikelupo:
Thanks for this commit.

PR #769, opened 2016-01-18
Properly compute header height when media picker is within another view

bagage:
Hardcoding header height is obviously not working when media picker view is embedded in another view. This PR does that automatically, but please confirm that it's the correct way of doing it... thanks!

### DIFF
--- a/Classes/KIFUITestActor.m
+++ b/Classes/KIFUITestActor.m
@@ -726,15 +726,20 @@
     
     // Wait for media picker view controller to be pushed.
     [self waitForTimeInterval:1];
-    
+   
     // Tap the desired photo in the grid
     // TODO: This currently only works for the first page of photos. It should scroll appropriately at some point.
-    const CGFloat headerHeight = 64.0;
+     UIAccessibilityElement *headerElt = [[UIApplication sharedApplication] accessibilityElementMatchingBlock:^(UIAccessibilityElement *element) {
+        return [NSStringFromClass(element.class) isEqual:@"UINavigationItemButtonView"];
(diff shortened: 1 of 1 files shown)


phatmann:
I am glad to see a hard-coded value go away. However, your calculation is not setting the header height, but rather its bottom offset. How about renaming `headerHeight` to `headerBottom`?

Note that you made your changes in your master branch instead of a new branch. If you amend your last commit with the change above and do a force push, then this PR will be updated. If you make a new commit you will need to squash them together before I can merge.

bagage:
Thanks for your fast comments @phatmann. I renamed `headerHeight` to `headerBottom` indeed. However I'm not sure what to do about your squash advice; I only have one commit different from your master, so it should be already ready to be merged, should not it be?

phatmann:
The commits look good. However, now I have another question (sorry I did not think of this 10 mins ago): if the system language is not English, the accessibility label will not be "Photos". Perhaps there is another way to locate the header view?

bagage:
Indeed, I'm running only English version because of such constraints... I'll try to find another way of finding it, maybe by searching views of class `UINavigationItemButtonView`..? Any hint is welcome!

bagage:
I updated PR by looking for the `UINavigationItemButtonView` element of the current page. Obviously if there are 2 it might return the invalid one, but I do not think that it's likely to happen. What do you think?

phatmann:
@bagage: this seems like a good approach, and is certainly better than hard-coding the English string or the height. Will merge. Thanks for your work on this.

PR #771, opened 2016-01-20
Fix search bar test to explicitly set autocapitalizationType

### LINKED ISSUE #757: enterText:intoViewWithAccessibilityLabel: could not enter the first character as lower case under auto-capitalization
It seemed that the shift key was not checked or not toggled before entering the first character. In other iOS and KIF versions, the first character was entered properly.

Simulator 9.2 as iPhone 6s
iOS 9.2
KIF v3.3.1

In the test,
[tester enterText:@"example" intoViewWithAccessibilityLabel:@"Title"];

In the console,
The step timed out after 1.00 seconds: Failed to get text "example" in field; instead, it was "Example"

justinseanmartin:
Resolves #757

### DIFF
--- a/Test Host/TableViewController.m
+++ b/Test Host/TableViewController.m
@@ -8,6 +8,8 @@
 
 @interface TableViewController : UITableViewController
 
+@property (weak, nonatomic) IBOutlet UISearchBar *searchBar;
+
 @end
 
 @implementation TableViewController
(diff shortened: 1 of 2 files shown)


justinseanmartin:
Any idea what's up with Travis? @bnickel @phatmann @mikelupo

mikelupo:
One of the jobs fail due to two flakey tests but the other job passed.

justinseanmartin:
Fixed the flakey tests in #773

PR #772, opened 2016-01-20
Fix modal tests to pass on iPads

justinseanmartin:
CC: @plu


justinseanmartin:
You can see this failure in action here:
https://travis-ci.org/kif-framework/KIF/jobs/103951800

PR #773, opened 2016-01-21
Fix tableview tests (causing Travis to be red)

justinseanmartin:
It looks like the accessibility behavior on iOS 9.0 had a bug. I tried locally and could reproduce against the 9.0 simulator, but not against the 9.1 or 9.2 simulator. I've adjusted the app behavior to account for the platform bug.


phatmann:
Nice fix. Will merge. One question: The `nullable` keyboard will prevent this code from compiling on older versions of Xcode -- should we care?

justinseanmartin:
Ooh, yes, sorry, didn't think about that, I should just drop them.

PR #774, opened 2016-01-21
Set default test configuration to Debug from Coverage

justinseanmartin:
I don't believe that anyone is using the files for anything useful at the moment and it makes a bunch of noise in the console logs when running tests. I switched the default project config to be `Debug` and deleted the checked in gcda file.


phatmann:
@bnickel: are you ok with this change?

justinseanmartin:
There may be some other place where this is being used that I didn't realize. From my perspective, it just generates noise in the console logs and dirty files on the the file system when running tests.

justinseanmartin:
Ok to merge?

mikelupo:
I think we're waiting to hear from @bnickel on this change

phatmann:
I will merge the change since we have not heard any objections from @bnickel. We can always revisit this as appropriate.

PR #775, opened 2016-01-21
Don't run acknowledgeSystemAlert tests on iOS7, it doesn't work

justinseanmartin:
I see the console logs with "AX Lookup problem! 22 com.apple.iphone.axserver:-1" against the iOS7 simulator. I'm running Xcode 6.4 on Yosemite against the 7.1 simulator.

Did this ever work on iOS7 simulators? Maybe we should do something to prevent it from being used on iOS7, like assert in that method itself? Would that make the contract more clear?


justinseanmartin:
Rebased with a green Travis

PR #776, opened 2016-01-21
Make gesture tests more reliable

justinseanmartin:
Some tests would fail when run individually (when the navigation controller push animation is still happening), fixed that with a sleep in the beforeAll.

Also, `testScrolling` was failing consistently on the iPhone 6 Plus. It was dragging from the left edge of the screen which was triggering the navigation controller's `interactivePopGestureRecognizer`. The edge swipe region seems significantly bigger on iPhone 6 Plus.


phatmann:
I love it, let's get the tests to be more reliable!

PR #779, opened 2016-01-21
Add a convenience method for logging an error

justinseanmartin:
We're using this as a convenience method to generate screenshots when test cases fail. It seemed like something that was reasonable to push upstream, but happy to keep it in our code if it wouldn't be useful for others.


phatmann:
Seems useful, and the name makes it clear what it does.

Issue #789, opened 2016-01-28
waitForAnimationsToFinish doesn't wait for animations on sublayers of views


rciobanu-omnifone:
Hi everyone,
I've encountered this issue (which I believe to be a bug) when using the `SVProgressHUD` cocoapod. The way that component works is by adding a sublayer to the HUD's layer and animating that sublayer.
Although I have my gripes with `SVProgressHUD` and spinners in general, I think in this particular case it's a `KIF` shortcoming as a user would see the HUD as animating and would wait for it to go away before progressing. Obviously when running tests this can lead to actions being attempted when the VC isn't in a ready state yet.
I've tracked the issue to the `- (void)waitForAnimationsToFinishWithTimeout:(NSTimeInterval)timeout` method and particularly line 153:
```obj-c
BOOL hasAnimation = view.layer.animationKeys.count != 0 && ![view.layer.animationKeys isEqualToArray:@[@"_UIParallaxMotionEffect"]];
```
I am ready to submit to your consideration a pull request that fixes the issue.

phatmann:
Fixed in the PR

PR #790, opened 2016-01-28
Fix for issue #789: waitForAnimationsToFinish animations on sublayers

rciobanu-omnifone:
The fix goes recursively through all sublayers of the current view's layer.
I've implemented the fix as a block to keep the changes to a minimum but this can easily be changed to a method, perhaps in a similar fashion to
```obj-c
- (void)performBlockOnDescendentViews:(void (^)(UIView *view, BOOL *stop))block
```

Comments are welcome.

### DIFF
--- a/Classes/KIFUITestActor.m
+++ b/Classes/KIFUITestActor.m
@@ -12,6 +12,7 @@
 #import "UIWindow-KIFAdditions.h"
 #import "UIAccessibilityElement-KIFAdditions.h"
 #import "UIView-KIFAdditions.h"
+#import "CALayer-KIFAdditions.h"
 #import "UITableView-KIFAdditions.h"
 #import "CGGeometry-KIFAdditions.h"
 #import "NSError-KIFAdditions.h"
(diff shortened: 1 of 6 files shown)


Classes/KIFUITestActor.m
```diff
@@ -149,8 +149,26 @@ - (void)waitForAnimationsToFinishWithTimeout:(NSTimeInterval)timeout {
         __block BOOL runningAnimationFound = false;
         for (UIWindow *window in [UIApplication sharedApplication].windowsWithKeyWindow) {
             [window performBlockOnDescendentViews:^(UIView *view, BOOL *stop) {
-                BOOL isViewVisible = [view isVisibleInViewHierarchy];   // do not wait for animatinos of views that aren't visible
-                BOOL hasAnimation = view.layer.animationKeys.count != 0 && ![view.layer.animationKeys isEqualToArray:@[@"_UIParallaxMotionEffect"]];    // explicitly exclude _UIParallaxMotionEffect as it is used in alertviews, and we don't want every alertview to be paused
+                BOOL isViewVisible = [view isVisibleInViewHierarchy];   // do not wait for animations of views that aren't visible
+                BOOL(^layerAnimationTest)(CALayer *layer);
+                __block __weak BOOL(^weakLayerAnimationTest)(CALayer *layer); // There was no retain cycle anyway, but this is the easiest way to keep the compiler happy
+                layerAnimationTest = ^(CALayer *layer) {
```
phatmann: Instead of using a block, perhaps it would be better to extend CALayer with a `hasAnimations` method.

rciobanu-omnifone: I agree completely. Will do.

rciobanu-omnifone:
I'm confused about Travis failing the compilation. Could you please re-run the tests?

phatmann:
This is a good PR, thanks for putting it together and getting Travis to pass! A few requests before I merge it:

* Please squash the 5 commits. Since you are using your master branch, please make sure no extra commits snuck in.
* Do we have a new dependency on QuartzCore? If so, we need to update the README and the Podspec file.

rciobanu-omnifone:
Done. Thanks for your patience and assistance in dealing with this PR.

PR #795, opened 2016-02-02
Fix for offscreen tableviewcell content views with gesture recognizers

justinseanmartin:
We had an issue where an offscreen element (TableViewCell content view) was being returned as a match for tapping, even though it wasn't visible on the screen. It would fail with a "View is not tappable" error.

The problem required changing several pieces of behavior in UIView-KIFAdditions:
1. Any element with a tap gesture recognizer was considered "tappable", even if no portion of it was visible on screen.
2. Prefer finding an element in a tableView that can be scrolled into view over an offscreen matching element that can't be tapped.
3. Table Views have cells that aren't in the visible region of the screen, but are still within the set of visibleCells. These shouldn't be skipped as part of `indexPathsForVisibleRows`.

The change is easiest to diff if you ignore whitespace in the diff (e.g. https://github.com/kif-framework/KIF/pull/795/files?w=1).

I ran our full automation suite and everything seems to work fine with the change. I'm going to let Travis do the heavy lifting for the KIF test suites.

### DIFF
--- a/Additions/UIView-KIFAdditions.m
+++ b/Additions/UIView-KIFAdditions.m
@@ -703,8 +703,27 @@ NS_INLINE BOOL StringsMatchExceptLineBreaks(NSString *expected, NSString *actual
             }
         }
     }];
-    
-    return hasTapGestureRecognizer;
+
+    if (!hasTapGestureRecognizer) {
+        return NO;
+    }
+
+    // Tap gesture recognizers are only useful if the elements are visible within their ancestor views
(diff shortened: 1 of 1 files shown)


justinseanmartin:
This isn't quite ready to merge yet, still doing some experiments.

justinseanmartin:
Going to close this for now, instead of leaving it open indefinitely. I'll come back to this when I have time to fix the issues that this breaks in our app. That said, I think the breakages in our app are because we were doing some bad/wrong things in a few places with table views.

PR #794, opened 2016-02-02
Add a new test actor, KIFUIViewTestActor.

RoyalPineapple:
add a new test actor, KIFUIViewTestActor, which separates the APIs of finding from acting on an element.

I submitted this as a PR in 2015 and we decided that it would be added in KIF v.4 but we'd need to work out a deprecation plan that doesn't disrupt testing for consumers of KIF.  Since then we've been using this API internally at Square and have come up with a two phase deprecation plan to get this merged. 

I've outlined the migration proposal here, feel free to suggest any changes or additions. 
https://docs.google.com/document/d/1cD_ymYtSBQ958e5QFOM-PdwvWMHhuafjt-rpGXM5V8M/pub

We'd like to get these changes merged soon as one of our goals right now is to move off our internal fork of KIF and start using the public version. (you've noticed all the PRs from Justin and me)

PR #800, opened 2016-02-23
Fixing acknowledgeSystemAlert for dirty simulators.

RoyalPineapple:
Because the system will only pop the system alerts once per simulator, we cant assert that they will pop on every test run. This PR fixes acknowledgeSystemAlert to ensure the alert is valid before we attempt to tap the dismissal button (avoiding a crasher), and updates the associated tests to only assert that the _final_ alert dismissal returns NO.

### DIFF
--- a/Classes/UIAutomationHelper.h
+++ b/Classes/UIAutomationHelper.h
@@ -8,6 +8,7 @@
 
 #import <Foundation/Foundation.h>
 
+
 @class KIFTestActor;
 
 @interface UIAutomationHelper : NSObject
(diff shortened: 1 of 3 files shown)


RoyalPineapple:
This fixes the travis test failure for this PR https://github.com/kif-framework/KIF/pull/794

justinseanmartin:
A few places I'd use dot syntax `alert.buttons lastObject` and `alert.isValid`, but generally LGTM assuming a green light from Travis.

phatmann:
Travis passed in all configs except for iOS 8, and that failure seemed to be test env related (testWait waited too long). So I am going to merge this.

I agree using dot syntax would be better for all property access, but we can do another PR for that later if desired. I don't think the code base is wildly consistent in this regard anyway :-)

PR #801, opened 2016-02-24
Define `kKIFMajorSwipeDisplacement` proportionally to screen width

### LINKED ISSUE #601: Swiping and UIPageViewController
Swiping through views that belong to a UIPageViewController doesn't seem to work properly. Sometimes it goes correctly to the next or previous page (depending on the swipe direction), but often it jumps over a page or stays on the same page.

I think the problem lies in ```viewContainingAccessibilityElement:tappable:error:``` in the UIAccessibilityElement-KIFAdditions category. It calls ```[UIScrollView scrollViewToVisible]```but since the UIPageViewController stacks three view controllers (it pre-renders the previous and next page) the ```viewContainingAccessibilityElement``` method seems to get hold of the wrong one and scrolls that to be visible. I'm not completely sure if this is the problem though, maybe I can find some time to dig into this a bit further the next days...

srking:
Value is set so that the previous value of 200pts is maintained for screens that are 320pts wide.

Fix #601

### DIFF
--- a/Additions/UIScreen+KIFAdditions.h
+++ b/Additions/UIScreen+KIFAdditions.h
@@ -10,4 +10,6 @@
 
 @interface UIScreen (KIFAdditions)
 
+@property (nonatomic, readonly) CGFloat majorSwipeDisplacement;
+
 @end
--- a/Additions/UIScreen+KIFAdditions.h
+++ b/Additions/UIScreen+KIFAdditions.h
@@ -1,0 +1,13 @@
+//
+//  UIScreen+KIFAdditions.h
+//  KIF
+//
+//  Created by Steven King on 25/02/2016.
+//
+//
+
+#import <UIKit/UIKit.h>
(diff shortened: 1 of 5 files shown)


phatmann:
I like the idea behind this change, but placing code in a #define macro can cause issues. I suggest instead that we get rid of the macro and replace it with a method.

srking:
I've failed to replicate the failing test (`testAuthorizingLocationServicesAndNotificationsScheduling`) locally. Is that test known to be unreliable?

mikelupo:
No, I have kicked off that particular job again. Let's see if it fails, and if so, the same way as prior.

phatmann:
Travis passed

srking:
@phatmann would you like any further changes to be made before this pull request can be accepted?

phatmann:
Now that macro is gone I merged it.

PR #805, opened 2016-03-02
CocoaPods 1.0 Migration

wanbok:
`exclusive` will be removed from [CocoaPods 1.0 Migration](http://blog.cocoapods.org/CocoaPods-1.0/).


phatmann:
Nice catch, thanks! It is interesting to see how much is changing in CocoaPods 1.0.

PR #810, opened 2016-03-08
Update headers

RoyalPineapple:


justinseanmartin:
LGTM

phatmann:
Thanks for fixing this up. Will push another version to CocoaPods now.

phatmann:
And great idea having Travis build the framework, so this never happens again.

PR #811, opened 2016-03-08
Fix Travis script to run both configs

justinseanmartin:
The exit on the end of the first command prevented the second one from running.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,4 @@ install:
   - if [ -z "$(brew ls --versions coreutils)" ] ; then brew install coreutils ; fi
   - gem install xcpretty --no-rdoc --no-ri --no-document --quiet
 script:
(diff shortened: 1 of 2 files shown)


phatmann:
All of the Travis runs are now failing with:

```
The command "/bin/bash -c "NSUnbufferedIO=YES xcodebuild test -scheme KIF -derivedDataPath=${PWD}/build/KIF -destination 'platform=iOS Simulator,${SIMULATOR}' | xcpretty -c && exit ${PIPESTATUS[0]}"" exited with 1.
```

justinseanmartin:
Yea, I saw that. I'm going to make it a single shell script that is checked in and call out to that.

justinseanmartin:
No worries, I just noticed when I wanted to see how much additional time it takes to do the extra build. Turns out it adds ~2 minutes to the overall CI response time, which isn't too bad given that it was around ~19 minutes.

Issue #243, opened 2013-08-27
Option to use accessabilityIdentifier instead of accessibilityLabel


grgcombs:
Given that `accessibilityLabel` is an outwardly-facing string that is actually used by accessibility screen readers (and should be localized to the device user's language), Apple now provides an alternate property (iOS 5+) that is specifically intended for UI Automation purposes:

``` Objective-C
@property(nonatomic, copy) NSString *accessibilityIdentifier;
```

> An identifier can be used to uniquely identify an element in the scripts you write using the UI Automation interfaces. Using an identifier allows you to avoid inappropriately setting or accessing an element’s accessibility label.
> 
> **Availability**
> Available in iOS 5.0 and later.
> 
> **_Source:**_ _[UIAccessibilityIdentification Protocol Reference](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIAccessibilityIdentification_Protocol/Introduction/Introduction.html)_

Ideally, the KIF framework would give us an easy means to swap out the logic that looks for `accessibilityLabel` and instead use `accessibilityIdentifier`.

PR #812, opened 2016-03-09
Move IdentifierTests/ under Classes/ so that it gets picked up by the podspec

Ushhud:
Moving IdentifierTests/ under Classes/ so that it gets picked up by the podspec (in order to make it available via pods).


justinseanmartin:
It is currently available in the subspec "IdentifierTests" if you pull that in. Please file an issue and we can discuss making changes to the design. If we do make this change, we'd likely also remove the existing subspec as well, which is a breaking change.

justinseanmartin:
Also, the reasoning is that AccessibilityIdentifier usage should be discouraged in general, because it doesn't make the app more accessible to real users using Voice Over. The issue has been discussed several times in existing issues, such as:

https://github.com/kif-framework/KIF/issues/243

Ushhud:
Oops, sorry missed that subspec. Thanks!

PR #813, opened 2016-03-10
fixed a bug where matching predicates could end up with an empty erro…

RoyalPineapple:
…r discription if subsequent predicates fail


RoyalPineapple:
@justinseanmartin pointed this bug out to me.

phatmann:
Merging, 3/4 Travis runs passed.

PR #816, opened 2016-03-15
fixed a bug where kif is unable to find the view for a UITableViewLabel

RoyalPineapple:


justinseanmartin:
LGTM

RoyalPineapple:
@phatmann @mikelupo I'd love another pair of eyes on this before a merge. 
 Additionally, would either of you mind adding @justinseanmartin and Me to the cocoapods owners list.

phatmann:
@RoyalPineapple: to make you and Justin CocoaPods owners I will need you to register with CocoaPods and then tell me your emails.

phatmann:
This code certainly does what it is advertised to do. I cannot speak to whether it is effective, but I assume you have empirical proof that it is. It would be good to have a test that proves it, but I am fine merging without a test because the change is localized and safe.

mikelupo:
My vote is to have a test case to prove that the code is working as advertised. 
Just for a little insurance 
:)

RoyalPineapple:
@mikelupo  I agree adding a test case would be good. but this is kinda hard to replicate. I have a reproducible case in our app (Where we're doing some crazy table view stuff) and have confirmed that this fixes the issue.
 in most circumstances this would be pretty benign as a UITableViewLabel only exists on cells created using the apple default tableViewCell styles, not on custom views.  

I'm doing some testing to ensure this doesn't cause any issues,

RoyalPineapple:
After digging into this further (I'm deep in UITableViewAccessibility) I've decided to decline this PR and come up with a more targeted approach. apologies

Issue #596, opened 2015-03-13
tapViewWithAccessibilityLabel fails when accessibilityLabel is not specifically set


SimonasA:
AccessibilityLabels works out of the box in iOS: you set title for UIButton and voice over uses that buttons title as a accessibilityLabel if you specifically do not set accesibilityLabel for that UIButton.

I would love to avoid setting accesibilityLabel's for all tested views as it would look more or less like this in all places:

```
[button setTitle:text forState:UIControlStateNormal];
button.accessibilityLabel = text;
```

I mean, - I would love not to dublicate and avoid setting accessibilityLabel for my buttons.

KIF works just fine finding those views with only titles set:
```
[tester waitForViewWithAccessibilityLabel:text];
```

but it fails when it is time to tap those views:
```
[tester tapViewWithAccessibilityLabel:text];
```

The thing is (my guess),- it finds UIButtonLabel (private apple label) as a view with that accessibility label instead of intentional button and fails to tap it.

Is KIF not working with out-of-the box accesibility labels intentionally and by design?
Or that is a bug?

(Sorry if it was already discussed,- checked all open/closed issues,- did not find anything similar)

Thanks

jmig:
For me it works out of the box 99% of the time.

Can you try to use :
````
[tester waitForTappableViewWithAccessibilityLabel:text traits:UIAccessibilityTraitButton];
//And then
[test tapViewWithAccessibilityLabel:text traits:UIAccessibilityTraitButton];
````

phatmann:
@SimonasA, does the solution proposed by @jmig work for you?

SimonasA:
So I updated POD, ran tests again and it works as expected (even without specifically setting ```UIAccessibilityTraitButton```)
I checked what has changed in KIF code and found nothing, so I couldnt say I am fully satisfied :}}
As ir works now, I am closing the issue. 

Guys, thank you very much for your help and attention! ; )

Benuuu:
I'm noticing an issue with this as well, but only specifically on the first launch after a simulator "Resent Contents and Settings".  Looking at the debugger it appears that my UILabels have the text set, but inspecting their accessibilityLabel shows `nil`.

Tested with Xcode 6.3.2 iOS 7.1 simulator against KIF 431a9f11e1e342ec6b27a172742d6e2c15b01532

mikelupo:
What about this commit? Does it still happen there?
[...]

PR #819, opened 2016-03-23
Fixed XCTest load order issues exposed in Xcode 7.3

benasher44:
It turns out that if you don't ensure that `XCTestObservationCenter` has already loaded, then attempting to use it and add an observer subverts some sort of internal setup required to get `XCTest` to spit out log lines during your tests like this:

> Test Case -[MyTestTarget.testOne testMethodOne] started.

This is a really big problem if you rely on something like `xcpretty` parsing this output from `XCTest` to later generate a test report. To fix this, I moved any loading that `KIF` needed from a `+(void)load` method, into `+[KIFTestCase setUp]`. This fixed the issue, and now all of our important `XCTest` logging is back.

Things I tried that didn't work:
1. Doing the setup in `+[XCTestObservationCenter initialize]` (not a good idea for other reasons, but I tried anyway)
2. Doing the setup in a function with `__attribute__(constructor)`
3. Moving the setup into `+(void)load` on `XCTestObservationCenter`. This appeared to work initially, but  this didn't work when switching from using KIF as a static library to using it as a dynamic framework. So, I added a `dispatch_async`, and this appeared to work in every case. But, I found with further testing that enabling accessibility here wasn't nearly as reliable as enabling it in the `XCTestObservation` method (as it was previously).

Here are the docs on `+(void)load` for reference:
>
Invoked whenever a class or category is added to the Objective-C runtime; implement this method to perform class-specific behavior upon loading.
>
>
The load message is sent to classes and categories that are both dynamically loaded and statically linked, but only if the newly loaded class or category implements a method that can respond.
The order of initialization is as follows:
>
>
1. All initializers in any framework you link to.
2. All +load methods in your image.
3. All C++ static initializers and C/C++ `__attribute__(constructor)` functions in your image.
4. All initializers in frameworks that link to you.
>
In addition:
>
- A class’s `+load` method is called after all of its superclasses’ `+load` methods.
- A category `+load` method is called after the class’s own `+load` method.
In a custom implementation of load you can therefore safely message other unrelated classes from the same image, but any load methods implemented by those classes may not have run yet.
> 

I also went ahead and applied similar treatment I initially applied to the `KIFAccessibilityEnabler` setup, wherein I moved it into the `+ (void)load` method on `XCTestObservationCenter` because that's the class it needed to use (i.e. attempt 3 above), to the `UIApplication` swizzling as well. In this case, we'd also want to also be sure that the `UIApplication` class is loaded in the runtime before interacting with it.

### DIFF
--- a/Classes/KIFAccessibilityEnabler.h
+++ b/Classes/KIFAccessibilityEnabler.h
@@ -10,4 +10,7 @@
 
 @interface KIFAccessibilityEnabler : NSObject
 
++ (instancetype)sharedAccessibilityEnabler;
+- (void)enableAccessibility;
+
 @end
--- a/Classes/KIFAccessibilityEnabler.m
+++ b/Classes/KIFAccessibilityEnabler.m
@@ -11,24 +11,13 @@
 #import <dlfcn.h>
 
 
-#if __IPHONE_OS_VERSION_MAX_ALLOWED < 90000
-@protocol XCTestObservation <NSObject>
-@end
-
-@interface XCTestObservationCenter : NSObject
-+ (XCTestObservationCenter *)sharedTestObservationCenter;
-- (void)addTestObserver:(id <XCTestObservation>)testObserver;
-@end
-#endif
-
-
 @interface AccessibilitySettingsController
 - (void)setAXInspectorEnabled:(NSNumber*)enabled specifier:(id)specifier;
 - (NSNumber *)AXInspectorEnabled:(id)specifier;
 @end
 
 
-@interface KIFAccessibilityEnabler () <XCTestObservation>
+@interface KIFAccessibilityEnabler ()
 
 @property (nonatomic, strong) id axSettingPrefController;
 @property (nonatomic, strong) NSNumber *initialAccessibilityInspectorSetting;
@@ -37,18 +26,6 @@
 
 
 @implementation KIFAccessibilityEnabler
-
-+ (void)load
-{
-    @autoreleasepool {
-        if ([XCTestObservationCenter respondsToSelector:@selector(sharedTestObservationCenter)]) {
-            XCTestObservationCenter *observationCenter = [XCTestObservationCenter sharedTestObservationCenter];
-            [observationCenter addTestObserver:[self sharedAccessibilityEnabler]];
-        } else {
-            [[self sharedAccessibilityEnabler] _enableAccessibility];
-        }
-    }
-}
 
 + (instancetype)sharedAccessibilityEnabler
 {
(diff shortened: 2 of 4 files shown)


benasher44:
@schluete here's the fix

phatmann:
Thanks for the experimentation and the fix. If someone in the future finds a way to fix this without the `dispatch_async` hack, then we can use that. But at least we have a fix for now.

benasher44:
@phatmann may want to hold off on merging this just yet, as we're still running this against our internal suite to see if other issues crop up. Will update soon once that's done, but I thought I'd just go ahead and get this out there

phatmann:
Good, coz it makes me nervous to have race condition hacks :-)

benasher44:
@phatmann okay so on our end things look good, but I wanna talk about this some more (because the hack and travis is failing). Is there any reason why we need to do it this way (using `XCTestObservationCenter`)? It seems like we could just go ahead and call `_enableAccessibility` once in `+ (void)load` like how it was done previously in the case that `XCTestObservationCenter` isn't available). This seems sufficient to me, since with `XCTestObservationCenter` we're setting up when we're told that the test bundle is starting and resetting when we're told the test bundle is finished. When the test bundle finishes, the program exits, so it seems like we could just set this up as soon as we know `KIFAccessibilityEnabler` and not worry about teardown. Then, we avoid this issue altogether.

Is there something I'm missing?

phatmann:
The teardown is done to restore the simulator to its initial state. See https://github.com/kif-framework/KIF/commit/927050d0a3e4178182570c560f646fa8a257bc52.

benasher44:
@phatmann Ah okay. How about we just do the setup in `+ (void)load`, and then I'll add a destructor function like so:
```
__attribute__((destructor))
void ResetAccessibilityInpector() {
  [[KIFAccessibilityEnabler sharedAccessibilityEnabler] _resetAccessibilityInspector];
}
```
I've confirmed that this does get called write before the test bundle quits.

benasher44:
Okay, I looked back at #596, which appears to be the discussion that led to this change, and after that and some experimentation, it's pretty clear that all of this behaves far more reliably if we rely on `XCTestObservationCenter`. So, I'm back in the camp of keeping this hack.

benasher44:
Actually, doing this in `+setUp` for KIFTestCase seems to work pretty well. Going to see what travis thinks.

benasher44:
Okay, see the comment about why I now can't use `XCTestObservation` for cleanup.

On the plus side, this does make `KIFAccessibilityEnabler` more flexible: even if you're not using XCTestCase (is this true for some people? because this class is written like that's true), you have an API to use `KIFAccessibilityEnabler` yourself, and KIF will guarantee the cleanup for you regardless.

benasher44:
Alrighty this looks good on our end, and now it's passing travis! :)

phatmann:
Looks like a good set of changes. Can you please squash the commits? Then I can merge it.

benasher44:
Done!

PR #821, opened 2016-03-25
Fix logical not precedence error. Fixes #817

### LINKED ISSUE #817: Building KIF with XCode Version 7.3 beta (7D162j) errors on logical not case
2016-03-21 at 11 58 am

These two lines in TappingTest.m are highlighted as errors:

* https://github.com/kif-framework/KIF/blob/master/KIF%20Tests/TappingTests.m#L77
* https://github.com/kif-framework/KIF/blob/master/KIF%20Tests/TappingTests.m#L95

I'm an obj-c newbie but [according what I can find on Operator Precedence](http://www.techotopia.com/index.php/Objective-C_2.0_Operator_Precedence), this error is legitimate: the line

```objc
if (! newValue == (originalValue -1))
```

Will be interpreted as

```objc
if ((! newValue) == (originalValue -1))
```

Because ! has higher precedence than ==, so the comparison will be

```objc
if (NO == (originalValue -1))
```

Which I assume is incorrect. Happy to write a PR if this is the case.

tmcw:


phatmann:
Let's use `!=` operator for clarity.

tmcw:
Cool, done!

PR #825, opened 2016-03-29
added all files to the framework and switched import syntax

RoyalPineapple:


brennantaylor:
Having some trouble using this as a framework. It looks like some headers got their attributes cleared. `KIFTestActor.h` is no longer part of the public headers for example.

I'm cleaning and trying again to make sure though.

RoyalPineapple:
@brennantaylor Whoops, thanks for noting that. this was a 3 minute pass as I was leaving the office. I'll have a deeper look and clean that up tomorrow.

RoyalPineapple:
@brennantaylor I cleaned this up a bit and was less aggressive with moving things around. I've got it building and running in my test environment. Can you try this branch again and see if it solves your issue?

brennantaylor:
@RoyalPineapple Any update?

RoyalPineapple:
@brennantaylor I pushed changes to the branch this afternoon and it's working in my environment. Can you give it a shot and report back?

Classes/KIFUIViewTestActor.m
```diff
@@ -6,11 +6,12 @@
 //
 //
 
-#import "KIFUIViewTestActor.h"
-#import "UIWindow-KIFAdditions.h"
-#import "NSPredicate+KIFAdditions.h"
-#import "UIAccessibilityElement-KIFAdditions.h"
-#import "NSString+KIFAdditions.h"
+#import <KIF/KIFUIViewTestActor.h>
```
justinseanmartin: I think that using quotes should be the norm for inclusions within source files. If this is generating an error, I think it is something we should take a look at.

justinseanmartin:
I rekicked the one red Travis build. The changes LGTM. Recommend that we add a test target that validates this doesn't regress in the future.

brennantaylor:
@RoyalPineapple This lgtm now.

phatmann:
Do we want to make a test target per @justinseanmartin's comment before we merge this?

RoyalPineapple:
@phatmann just added one, Lets see how Travis does.

brennantaylor:
That was quick. Thanks!

This is blocking our team from updating to xcode 7.3 so it would be nice if this were resolved soon. However I am willing to wait if that means this won't regress again.

scripts/ci.sh
```diff
@@ -19,6 +19,7 @@ rm -rf ${PWD}/build
 # Frameworks are only supported on iOS8 and later
 if [[ ! ${SIMULATOR} =~ .*OS=7.* ]]; then
   env NSUnbufferedIO=YES xcodebuild build -derivedDataPath=${PWD}/build/KIFFramework -scheme KIFFramework -destination "platform=iOS Simulator,${SIMULATOR}" | xcpretty -c
```
justinseanmartin: I'd just get rid of the building command, because this is implied behavior from the test command that happens directly after this.

phatmann:
Are we good to merge?

RoyalPineapple:
@phatmann travis was being flakey for one configuration so i babysat it yesterday, 3 of the last 4 runs were stable while one run hung on launching the sim (I'm betting thats just general travis flakiness). Im feeling comfortable merging this now.

phatmann:
Alright, merging now! This will solve a bunch of issues for folks, thanks for digging in and getting this done.

phatmann:
I pushed a new release to CocoaPods with this fix.

Issue #834, opened 2016-04-13
Rename system macro


kkodev:
**system** macro is way too common name and causes clashes with other libraries/code. I suggest renaming it to **kif_system**.

phatmann:
I would like to get rid of the macros entirely. If you have an idea, let us know.

kkodev:
Well, this (and **tester**) could just be convenience static initialisers in appropriate classes.

phatmann:
Sounds good to me. Would you consider modifying #835 to do this so we can all see how it looks and feels? As  long as it is performant and terse I imagine it will pass muster.

phatmann:
Also worth mentioning: for backwards compatibility we probably want to do the following:

1. In a new build, keep the macros around, but wrap them around the new methods. Warn users that macros are deprecated.
1. In a following build, move the macros into a separate header file that users can choose to include.

kkodev:
I will give it a go.
For new build perhaps the problem can be resolved with some #indef to allow to disable macros + warnings as you suggested.
Like #ifndef KIF_DISABLE_MACROS or similar.

bnickel:
I believe directives like that won't work if KIF is compiled as a framework and it would always be treated as though `KIF_DISABLE_MACROS` is undefined.

justinseanmartin:
Closing this, it is already covered by #733

PR #835, opened 2016-04-13
Rename system macro to kif_system

kkodev:
This addresses issue #834

### DIFF
--- a/Classes/KIFSystemTestActor.h
+++ b/Classes/KIFSystemTestActor.h
@@ -10,7 +10,7 @@
 #import "KIFTestActor.h"
 #import <UIKit/UIKit.h>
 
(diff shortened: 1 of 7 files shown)


mikelupo:
Two things. 
1) the change to fix the Operator in a test needs to be in a separate PR. Otherwise, this change looks fine.
2) I defer to @phatmann and @bnickel for the commit that renames system to kif_system.

kkodev:
Makes perfect sense. I extracted the operator change and will open separate PR

phatmann:
Obviously changing `system` to `kif_system` is a breaking change. If we are going to break clients, I would rather get rid of the macros entirely and use functions or methods instead. Then the pain of migrating is worth it.

justinseanmartin:
This is also being taken care of by #988

PR #836, opened 2016-04-13
Fixed operator precedence in a test

kkodev:


phatmann:
Please use the `!=` operator. The intent would be clearer. We fixed some other tests recently and that is what we did.

justinseanmartin:
Unchanged for over a year, and CI is green, so I think things are "fine" for the moment. Would be happy to discuss if someone wanted to revive the branch.

PR #838, opened 2016-04-16
Fixes various small documentation issues

steipete:
We compile with `-Wdocumentation` and found a few small issues while using KIF to test PSPDFKit.

Thanks for making this - it's very useful!


phatmann:
Thanks for the fixes.

steipete:
Thanks for merging! We have more incoming :)

PR #839, opened 2016-04-16
Add fix for a case where accessibility element frame was zero

steipete:
We ran into this and I wanted to open a PR in case anyone else sees this.

Probably misses a KIF test case and more details - ping me if you see or know about a similar issue.

### DIFF
--- a/Classes/KIFUITestActor.m
+++ b/Classes/KIFUITestActor.m
@@ -954,7 +954,14 @@
 
     // Within this method, all geometry is done in the coordinate system of the view to scroll.
 
-    CGRect elementFrame = [viewToScroll.windowOrIdentityWindow convertRect:element.accessibilityFrame toView:viewToScroll];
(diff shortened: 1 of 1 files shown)


phatmann:
I would like to understand what broke to cause this problem and in which testing scenarios you run into this issue. If we are working around a bug in UIAccessibility, then this is a fine fix, but if we are working around a bug in KIF itself, then we should fix the root cause.

steipete:
Agreed; need to do more digging here to see what broke. Will follow up.

fritzsche91:
As I have observed **similar issues** for srolling/swiping, this is potentially related to: https://github.com/kif-framework/KIF/pull/869

If you need any more information on this, feel free to contact me.

justinseanmartin:
The fix in #869 was merged and haven't heard other reports of this in over a year. Closing this issue out.

PR #840, opened 2016-04-16
Speed up wait time if layer speed is increased

steipete:
At PSPDFKit our UI tests start to get to a size where waiting for the KIF tests becomes impractical, so we investigated ways to make them run faster.

We had great success with increasing the layer speed on the key window:

```
UIApplication.sharedApplication.keyWindow.layer.speed = speed;
```

This might cause trouble for (very unusual) multi-window setups, but it's still entirely optional to use.
The default value here is 1, so no timeouts are changed by default.

Are there any other important spots where we should divide based on layer speed?

### DIFF
--- a/Additions/UIAccessibilityElement-KIFAdditions.m
+++ b/Additions/UIAccessibilityElement-KIFAdditions.m
@@ -144,7 +144,7 @@ MAKE_CATEGORIES_LOADABLE(UIAccessibilityElement_KIFAdditions)
             }
             
             // Give the scroll view a small amount of time to perform the scroll.
-            CFRunLoopRunInMode(kCFRunLoopDefaultMode, 0.3, false);
+            CFRunLoopRunInMode(kCFRunLoopDefaultMode, 0.3 / UIApplication.sharedApplication.keyWindow.layer.speed, false);
         }
         
         superview = superview.superview;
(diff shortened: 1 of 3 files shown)


phatmann:
This is a neat idea. Can you squash the three commits so I can merge it?

To your comment "Are there any other important spots where we should divide based on layer speed?": I would not be surprised if speeding up the keyWindow layer breaks KIF in other ways. But let's get this PR merged in so we can all play with it.

If this all works out well, we might want to expose the layer speed change via a KIF method to make this trick more obvious.

steipete:
Sure - will do! I have a blog post upcoming about this (including a very cool video on how we run our UI tests at 100x - ludicrous speed) and would love your feedback. Can give you preview access if you ping me at peter at pspdfkit com.

danydev:
@steipete hey Peter, do you think can you squash it soon? Would be great to have it merged

RoyalPineapple:
👍 looking forward to this. Loved the blog post.

steipete:
All squashed!

mikelupo:
A couple of the builds failed. I manually re-triggered them. If they pass I'll merge.

steipete:
Unless changed, `UIApplication.sharedApplication.keyWindow.layer.speed` should simply be 1 - so no real changes are in there. Is it possible that the tests are simply flaky?

mikelupo:
Of course it's possible. Actually the tests are OK, it's the travis infrastructure that I call to question. It's timing out waiting for the simulator to boot. I'll go ahead and merge. 
Thanks for your patience.

NachoSoto:
Unlikely to happen in any of these cases, but if there is no key window these would be dividing by 0 :P 
It's probably worth asserting that the layer isn't `nil`, otherwise it might be a hard bug to debug.

PR #849, opened 2016-05-02
Mark -[KIFTestActor init] unavailable.

dshusta:
Ran into a problem where we created a KIFTestActor without a line or file, and then when it ran into a test failure it would try and insert a `nil` for `self.file` into the exception user info in NSException+KIFAdditions and crash.

Elected to change KIFTestActor to simply not allow the user to create a KIFTestActor without a file path in the first place (this also works really well with Swift). 

An alternative solution is to perform a nil check before creating the dictionary in the NSException category.


phatmann:
Seems like a reasonable fix. Thanks for the PR.

PR #850, opened 2016-05-03
Additional speedups relative to animation speed

chillpop:
This builds on #840 and adds a more central way to scale animation wait times relative to animation speed.

- created `waitForTimeInterval:relativeToAnimationSpeed:` to make it explicit when times are scaled.
- created `KIFRunLoopRunInModeRelativeToAnimationSpeed` to do the same for `CFRunLoopRunInMode`
- added an `animationSpeed` property to UIApplication to simplify getting and setting the keyWindow animation speed (and add some safety to prevent divide by 0 errors)

All tests pass locally with the animation speed set to 100.  I left in the commented code I used to run the tests with fast animations for others to verify.  Ideally, I'd prefer to have a way to run the tests on Travis at fast animation speeds in addition to normal speed.  I'm open to suggestions for how best to implement that.

### DIFF
--- a/Additions/UIAccessibilityElement-KIFAdditions.m
+++ b/Additions/UIAccessibilityElement-KIFAdditions.m
@@ -144,7 +144,7 @@ MAKE_CATEGORIES_LOADABLE(UIAccessibilityElement_KIFAdditions)
             }
             
             // Give the scroll view a small amount of time to perform the scroll.
-            CFRunLoopRunInMode(kCFRunLoopDefaultMode, 0.3 / UIApplication.sharedApplication.keyWindow.layer.speed, false);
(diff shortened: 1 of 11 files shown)


phatmann:
I like these changes. Nice work.

Because you are using your master branch, the "fix compiler error" commit incorrectly ended up in this PR. The fix is incorrect, as it should be `newValue != oldValue` not `!newValue == oldValue`. Once the errant commit is removed I can merge this PR.

chillpop:
Thanks. I've removed the unrelated commit.

chillpop:
Any ideas on why the Xcode 7.0 / iOS 9.0 simulator keeps failing to start on Travis (and this failing one of the automated builds)?

phatmann:
Travis is flakey. If the iOS 8 tests are passing that is good enough for me.

PR #852, opened 2016-05-09
Fix for cell reuse causing wrong elements to be tapped

percysnoodle:
In our tests, we had a step which tapped a cell in a collection view, identified by its accessibility label.  When the test ran, the collection view was initially offscreen, and while it was being scrolled onto the screen the collection view decided to recycle its cells.  KIF had kept a reference to the original cell which it then used to send the tap event.  As a result, the wrong cell in the collection view was tapped, and the test failed.

This PR addresses the problem by re-checking the accessibility element after calling the method which can cause scrolling.  <s>The logic for identifying accessibility elements which was in the view category has been moved into the accessibility element category so that it can be used in both places.</s>

### DIFF
--- a/Additions/UIAccessibilityElement-KIFAdditions.m
+++ b/Additions/UIAccessibilityElement-KIFAdditions.m
@@ -50,6 +50,13 @@ MAKE_CATEGORIES_LOADABLE(UIAccessibilityElement_KIFAdditions)
         return NO;
     }
     
+    // viewContainingAccessibilityElement:.. can cause scrolling, which can cause cell reuse.
+    // If this happens, the element we kept a reference to might have been reconfigured, and a
+    // different element might be the one that matches.
+    if (![UIView accessibilityElement:element hasLabel:label accessibilityValue:value traits:traits]) {
+        return NO;
+    }
+    
     if (foundElement) { *foundElement = element; }
     if (foundView) { *foundView = view; }
     return YES;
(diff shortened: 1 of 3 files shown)


percysnoodle:
Ugh, looks like I misunderstood where UIAccessibilityElement sits in the class hierarchy.  I'll try again.

percysnoodle:
The CI failures don't make sense to me, so I'm going to close and re-open this to kick them off again.

phatmann:
All tests are passing for me locally. We need to make sure that the Travis error is not directly caused by this PR, because this change affects every KIF matching test.

percysnoodle:
All passing here too.  I don't have Xcode 7.0 or 7.2 to test in, though.

phatmann:
A few more questions...

With this PR we would be adding an extra test to *every single* KIF element match. What are the performance ramifications of this? Can we somehow limit this change to just affect UITableView and UICollectionView?

percysnoodle:
I did wonder about that - but to properly squash the bug, you'd need to check whether the view was a UI*ViewCell or contained inside one.  I'd have to profile it to be sure, but I imagine that walking up the view hierarchy for every test would be slower than doing an extra string comparison.

phatmann:
As long as the extra code runs fast enough we don’t need a special case. One could argue it makes KIF that much more robust.

justinseanmartin:
Looks like this change introduced the consistent failures we're seeing in Travis now. This is the first CI failure where the `PullToRefreshTests` started failed with the `BINGO!` error message.

phatmann:
@justinseanmartin: I can back out the errant PR now and then discuss with the author about taking another approach. Do you agree?

It seemed like an intrusive but safe change, but apparently I was wrong :-(

justinseanmartin:
We might as well put up a PR and see if things go back to mostly green with the change backed out. 

Done here: https://github.com/kif-framework/KIF/pull/880

PR #858, opened 2016-05-24
Use scroll view delegate to reliably wait for scrolling to finish

jmkk:
Instead of hard coded arbitrary waits, use scroll view proxy delegate to wait for scrolling to finish.
This removes the need to wait for an arbitrary amount of time (which may not be enough for slower machines, or may be too long for faster ones).
On a relatively new machine, this improves performance for lookups in scroll views (table views and collection views) by 2x and more.

### DIFF
--- a/Additions/UIAccessibilityElement-KIFAdditions.m
+++ b/Additions/UIAccessibilityElement-KIFAdditions.m
@@ -15,6 +15,8 @@
 #import "UIView-KIFAdditions.h"
 #import "LoadableCategory.h"
 #import "KIFTestActor.h"
+#import "KIFProxyDelegate.h"
+#import "KIFScrollViewDelegates.h"
 
 MAKE_CATEGORIES_LOADABLE(UIAccessibilityElement_KIFAdditions)
 
@@ -130,8 +132,17 @@ MAKE_CATEGORIES_LOADABLE(UIAccessibilityElement_KIFAdditions)
         if ([superview isKindOfClass:[UIScrollView class]]) {
             UIScrollView *scrollView = (UIScrollView *)superview;
             
+            // Replace the scroll view delegate to capture scrolling completion
+            KIFScrollViewDelegate * kifDelegate = [KIFScrollViewDelegate new];
+            KIFProxyDelegate * delegate = [[KIFProxyDelegate alloc] initWithOriginalDelegate:scrollView.delegate
+                                                                         replacementDelegate:kifDelegate];
+            scrollView.delegate = (id<UIScrollViewDelegate>)delegate;
+
             if (((UIAccessibilityElement *)view == element) && ![view isKindOfClass:[UITableViewCell class]]) {
+                // Initiate scrolling
                 [scrollView scrollViewToVisible:view animated:YES];
+                // Wait for scrolling to finish
+                [kifDelegate waitForScrollCompleteOnView:view inScrollView:scrollView];
             } else {
                 CGRect elementFrame = [view.window convertRect:element.accessibilityFrame toView:scrollView];
                 CGRect visibleRect = CGRectMake(scrollView.contentOffset.x, scrollView.contentOffset.y, CGRectGetWidth(scrollView.bounds), CGRectGetHeight(scrollView.bounds));
@@ -139,12 +150,15 @@ MAKE_CATEGORIES_LOADABLE(UIAccessibilityElement_KIFAdditions)
                 // Only call scrollRectToVisible if the element isn't already visible
                 // iOS 8 will sometimes incorrectly scroll table views so the element scrolls out of view
                 if (!CGRectContainsRect(visibleRect, elementFrame)) {
+                    // Initiate scrolling
                     [scrollView scrollRectToVisible:elementFrame animated:YES];
+                    // Wait for scrolling to finish
+                    [kifDelegate waitForScrollCompleteOnView:view inScrollView:scrollView];
                 }
             }
             
-            // Give the scroll view a small amount of time to perform the scroll.
-            KIFRunLoopRunInModeRelativeToAnimationSpeed(kCFRunLoopDefaultMode, 0.3, false);
+            // Restore the delegate to original
+            scrollView.delegate = delegate.original;
         }
         
         superview = superview.superview;
--- a/Additions/UIView-KIFAdditions.m
+++ b/Additions/UIView-KIFAdditions.m
@@ -14,6 +14,8 @@
 #import "UITouch-KIFAdditions.h"
 #import <objc/runtime.h>
 #import "UIEvent+KIFAdditions.h"
+#import "KIFProxyDelegate.h"
+#import "KIFScrollViewDelegates.h"
 
 double KIFDegreesToRadians(double deg) {
     return (deg) / 180.0 * M_PI;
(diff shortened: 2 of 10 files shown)


mikelupo:
@jmkk Thank you for your PR and Thank you for using a branch other than master to do your work on. 
One small request....could you please squash your commits into one?

This seems like kind of a risky PR. @bnickel, @phatmann can you please have look?

jmkk:
@mikelupo sure, once we've iterated over the PR (there might be more followup commits to it) I'll make sure to squash them into one.

There is some risk involved for existing projects, as previous explicit waits might have worked well for specific environments. Waiting for scroll views to tell us when scrolling has stopped improves both stability of the tests (doesn't make any assumptions about how long scrolling actually takes), and also improves performance (as in most cases we do not need to wait as long as previously hard-coded waits). We've used this logic for our main app with hundreds of tests with really good results. 

This is the first PR for KIF improvements we made at LinkedIn (see [this blog post](https://engineering.linkedin.com/blog/2016/01/ui-automation--keep-it-functional--and-stable-) for some details).

mikelupo:
Re: the blog. Great write up. Thanks for doing that.

I agree that your PR may expose the community to some unexpected failures because things run faster. 
While running tests on your branch, I made a small PR to see if I could stabilize some tests that ran particularly bad for me locally. I admit that I ran them on your branch for this PR but submitted my PR (PR#[https://github.com/kif-framework/KIF/pull/859]) against the current master. Anyhow, let's see if the Travis tests get any better. They've been a royal pain.

jmkk:
Actually, I would argue that with this PR the scrolling tests should become more stable.
"Things run faster" only when they can, and they run slow when the machine needs more time to perform and finish scrolling. With the previous approach, we had to tweak the delays in the scrolling code to achieve "stabilization" and hope that the scroll views indeed are ready to perform checks we wanted. 
With this change, we just know as the scroll view reports the fact that it has finished scrolling.

The only thing risky about this change is if some consumer out there started **depending** on the hard coded delays.

bnickel:
I'd be a little wary of changing the scroll view's delegate from within a test as setting the delegate could trigger side effects in the application.  Since you're only really interested in whether `scrollViewDidEndScrollingAnimation:` has been called and how long since its been since the last `scrollViewDidScroll:`, here are some approaches I'd consider.

**For `scrollViewDidEndScrollingAnimation:`**

`-[UIScrollView(UIScrollViewInternal) _delegateScrollViewAnimationEnded]` posts a notification named  "_UIScrollViewAnimationEndedNotification" whenever it calls `scrollViewDidEndScrollingAnimation:`, so you could just register for that.

**For `scrollViewDidScroll:`**

This call has high performance expectations, so UIKit doesn't post notification inside `-[UIScrollView _notifyDidScroll]`.  Instead it calls `_didScroll` and `_updateBackdropMaskViewsInScrollView:` on objects registered with `-[UIScrollView _addScrollNotificationView:]` and unregistered with `-[UIScrollView _removeScrollNotificationView:]`.

You could do something like:

```
@interface UIScrollView (Private)
- (void) _addScrollNotificationView:(id)obj;
- (void) _removeScrollNotificationView:(id)obj;
@end

@interface KIFScrollViewTracker : NSObject
@property (nonatomic, copy) NSDate *lastScrollTime;
@property (nonatomic, assign) BOOL didFinish;
@end

@implementation KIFScrollViewTracker
- (id)initWithScrollView:(UIScrollView *)scrollView
{
    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(done) name:"_UIScrollViewAnimationEndedNotification" object:scrollView];
    [scrollView _addScrollNotificationView:self];
    return self;
}

- (void)unregisterWithScrollView:(UIScrollView *)scrollView
{
    [scrollView _removeScrollNotificationView:self];
}

- (void)dealloc { [[NSNotificationCenter defaultCenter] removeObserver:self]; }
- (void)_didScroll { _lastScrollTime = [NSDate date]; }
- (void)_updateBackdropMaskViewsInScrollView:(id)obj {}
- (void)done { _didFinish = YES; }
@end

...

KIFScrollViewTracker *tracker = [[KIFScrollViewTracker alloc] initWithScrollView:scrollView];
/// SCROLL
do {
    /// SPIN
} while (!tracker.done && (tracker.lastScrollDate == nil ||  [NSDate timeIntervalSinceDate:tracker.lastScrollDate] < 0.1);
[tracker unregisterWithScrollView:scrollView];
```

The other option would be to swizzle `-[UIScrollView setContentOffset:]` or `-[UIScrollView _notifyDidScroll]` to capture the last scroll time for a given scroll view.

jmkk:
@bnickel Thanks, these are worth looking into. However, I would opt for a solution that doesn't depend on undocumented APIs wherever possible.

What side effects do you have in mind when using proxy delegate? The only thing I can think of is when the application ends up dynamically changing scroll view delegate, and it ends up happening mid-scroll, but I consider it to be asking for trouble (as the delegate could get only partially notified about what's going on with the scroll view). 

Any inputs would be appreciated - I'll be happy to change the implementation, as long as it avoids having hard coded delays and waits.

bnickel:
Depending on undocumented API's is pretty much KIF's thing. :)  In all seriousness though, that is a valid point.  The main concern I have is that scroll view subclasses tend to have non-trivial delegates and could do a lot of heavy lifting in `setDelegate:`.  I did a quick audit of `UIScrollView`, `UITableView`, and `UICollectionView` and the only concerning thing I see is this in `UITableView setDelegate:`

    if (!(BIT_TEST(*(r15 + rbx + 0xc), 0x26))) {
            r14 = [(r12)(rbx, @selector(_tableHeaderView:), 0x1, rcx) retain];
            r12 = (r12)(r14, @selector(isKindOfClass:), (r12)(@class(UISearchBar), @selector(class), 0x1, rcx), rcx);
            [r14 release];
            if (r12 != 0x0) {
                    [rbx setTableHeaderViewShouldAutoHide:0x1];
            }
    }
    r14 = var_40;
    r15 = var_48;
    if ((rbx->_rowData != 0x0) && (((var_51 ^ [rbx _estimatesRowHeights]) & 0xff) == 0x1)) {
            [rbx _updateRowData];
    }

It looks like in certain conditions, setting the delegate could cause the search bar to be hidden and when the table view is estimating row heights it causes some amount of work to be performed on the visible cells.  That could be an acceptable risk but it's at least worth thinking about.

The other way to observe last scroll time without depending on private APIs or delegates is swizzling.  You could still use `_UIScrollViewAnimationEndedNotification` notifications as that's not required to end scrolling in the current design, it just lets it end faster.

```
NSMapTable *KIF_lastScrollTimes;

@implementation UIScrollView (LastScrollTime)

+ (void)load
{
    Swizzle(self, @selector(setContentOffset:), @selector(KIF_setContentOffset:));
}

- (void)KIF_setContentOffset:(CGPoint)contentOffset
{
    if (KIF_lastScrollTimes == nil) {
        KIF_lastScrollTimes = [NSMapTable mapTableWithKeyOptions:NSMapTableWeakMemory valueOptions:NSMapTableStrongMemory];
    }
    [KIF_lastScrollTimes setObject:[NSDate date] forKey:self];
}

- (NSDate *)lastScrollTime
{
    return [KIF_lastScrollTimes objectForKey:self];
}

@end
```

jmkk:
FYI we're looking into the alternatives to minimize the risks. Stay tuned.

justinseanmartin:
Work on this PR appears to have been abandoned. Closing the PR, but feel free to reopen if work continues.

PR #859, opened 2016-05-24
Update Test cases for [hopefully] better travis results

mikelupo:
After running the test cases locally on both hardware and sim
it was noticed that some tests needed a bit more waiting to run
reliably. Perhaps running too fast on robust platform?

Notably:
Tests that type into input fields:
* In "(void)expectView:(UIView *)view toContainText:(NSString *)expectedResult"
the timeout was set to 1.0. This bumps it up to 2.0. Seems to stabilize field tests.
The community will also benefit from this change.
* Predictive text was getting in the way of typing the string, "Test-Driven Development".
The text was getting auto-corrected.  This PR changes it to "Testing"

Tests that utilize gestures:
* MultiFinger and TableView Tests benefit from waitFor... method calls.

### DIFF
--- a/Classes/KIFUITestActor.m
+++ b/Classes/KIFUITestActor.m
@@ -465,7 +465,7 @@
         KIFTestWaitCondition([actual isEqualToString:expected], error, @"Failed to get text \"%@\" in field; instead, it was \"%@\"", expected, actual);
         
         return KIFTestStepResultSuccess;
-    } timeout:1.0];
+    } timeout:[KIFTestActor defaultTimeout]];
 }
 
 - (void)clearTextFromFirstResponder
--- a/KIF Tests/TableViewTests_ViewTestActor.m
+++ b/KIF Tests/TableViewTests_ViewTestActor.m
@@ -109,6 +109,7 @@
     [[viewTester usingLabel:@"Edit"] tap];
 
     __KIFAssertEqualObjects([[viewTester usingIdentifier:@"TableView Tests Table"] waitForCellInTableViewAtIndexPath:[NSIndexPath indexPathForRow:-3 inSection:1]].textLabel.text, @"Cell 35", @"");
+	[viewTester waitForCellInTableViewAtIndexPath:[NSIndexPath indexPathForRow:-1 inSection:1]];
     __KIFAssertEqualObjects([[viewTester usingIdentifier:@"TableView Tests Table"] waitForCellInTableViewAtIndexPath:[NSIndexPath indexPathForRow:-1 inSection:1]].textLabel.text, @"Cell 37", @"");
 
     [[viewTester usingIdentifier:@"TableView Tests Table"] moveRowInTableViewAtIndexPath:[NSIndexPath indexPathForRow:-1 inSection:1] toIndexPath:[NSIndexPath indexPathForRow:-3 inSection:1]];
@@ -148,7 +149,7 @@
 
 - (void)testEnteringTextIntoATextFieldInATableCell
 {
-    [[viewTester usingLabel:@"TextField"] enterText:@"Test-Driven Development"];
+    [[viewTester usingLabel:@"TextField"] enterText:inputFieldTestString];
 }
 
 @end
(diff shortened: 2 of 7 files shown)


Classes/KIFUITestActor.m
```diff
@@ -444,7 +444,7 @@ - (void)expectView:(UIView *)view toContainText:(NSString *)expectedResult
         KIFTestWaitCondition([actual isEqualToString:expected], error, @"Failed to get text \"%@\" in field; instead, it was \"%@\"", expected, actual);
         
         return KIFTestStepResultSuccess;
-    } timeout:1.0];
+    } timeout:2.0];
```
jmkk: FYI in our code we bumped the timeout for text views/fields to 10s. Timeouts should be long (even 30s) to avoid flaky tests like it was the case before. On certain (slower) build machines and longer texts entered, even 2s might not be enough, and you'll be in the same place as before.

KIF Tests/TableViewTests_ViewTestActor.m
```diff
@@ -148,7 +149,7 @@ - (void)testButtonAbsentAfterSetHidden
 
 - (void)testEnteringTextIntoATextFieldInATableCell
 {
-    [[viewTester usingLabel:@"TextField"] enterText:@"Test-Driven Development"];
+    [[viewTester usingLabel:@"TextField"] enterText:@"Testing"];
```
jmkk: It might be a good idea to add a const for the `@"Testing"` string literal, and reference it from both places - with a comment explaining that it should not be changed to not mess with the auto-correct?

jmkk:
@mikelupo Thanks for fixing the tests! 

Note that the flakiness you've observed in the KIF test suite are caused by the same issues I pointed out in the [blog post](https://engineering.linkedin.com/blog/2016/01/ui-automation--keep-it-functional--and-stable-). It's not a good thing that the test suite for the test framework itself is not stable.

Adding more `waitForAnimationsToFinish` calls can temporarily alleviate the problem, but in the end it's not solving the root cause. "Stabilization wait" and related hard coded delays may work on some machines, but not all, and not always. 

Soon, I will be submitting PRs that helped us deal with this situation, and allowed us to remove all references to wait* APIs.

Classes/KIFUITestActor.m
justinseanmartin: I agree. I don't see why not just use the default 10s timeout here if a failure is going to get raised on timeout anyway.

justinseanmartin:
The text change as a workaround is good to get the tests stable, but ideally we should figure out how to avoid typing triggering the autocomplete.

mikelupo:
@justinseanmartin I agree. Sounds like a test case in and of itself. Maybe it was only that my device is the issue. Wondering if I should omit that part of the change. @jmkk I experienced the failures of those tests while running the tests against your PR. Was that expected?

jmkk:
@mikelupo text entry tests have nothing to do with my PR - it is flakiness. I've noticed such failures even without my changes - usually when the input locale on my machine was set to a non-EN keyboard. So I agree with the text string change.

Note that it is possible to turn off auto-complete on a text field (`setAutoCorrectionType`), so in theory we could do that just before entering text, and restore to what it was before after. It would test the text field in a bit different state than the prod environment though.

jmkk:
I just confirmed that running TypingTests (without your PR) with EN keyboard is passing, but with non-EN keyboard locale consistently fails.

justinseanmartin:
I think changing the text to something that works consistently is good, separate of having a more permanent fix for triggering autocorrection. 

I also agree that the more permanent solution probably looks like disabling autocorrect when typing, then restoring the state after typing is completed. Totally ok doing that outside the scope if these fixes though.

phatmann:
@mikelupo Do you still want to merge this based on the comments above?

phatmann:
@mikelupo before I merge can you respond to @jmkk comments and/or make the fixes he recommends?

mikelupo:
@phatmann @justinseanmartin @jmkk I have (finally) followed up with this PR.

phatmann:
@justinseanmartin @jmkk: all good?

phatmann:
Looks good. Rerunning Travis tests since they failed to run.

phatmann:
Travis is being flakey, but these changes look safe and good so I am going to merge.

PR #861, opened 2016-05-25
Added waitForDeleteStateForCell* methods to reliably wait for cell delete state changes

jmkk:
These APIs are useful to know when it's safe to perform delete operations when swiping left on table view cells in the edit mode.

### DIFF
--- a/Classes/KIFUITestActor.h
+++ b/Classes/KIFUITestActor.h
@@ -675,6 +675,18 @@ typedef NS_ENUM(NSUInteger, KIFPullToRefreshTiming) {
 */
 - (void)swipeRowAtIndexPath:(NSIndexPath *)indexPath inTableView:(UITableView *)tableView inDirection:(KIFSwipeDirection)direction;
 
+/*!
+ @abstract Waits for the given cell to transition to the delete state. Useful when swiping left on a cell for delete action.
+ @param cell Cell to wait for delete state on.
+ */
+- (void)waitForDeleteStateForCell:(UITableViewCell*)cell;
+
+/*!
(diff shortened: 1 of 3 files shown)


justinseanmartin:
FYI - @jmkk, it looks like this change introduced a regression in the iPad 2, iOS 7 test configuration if you have time to take a look.

jmkk:
Looking

phatmann:
@jmkk: any conclusions?

justinseanmartin:
Alternatively, we could potentially drop iOS 7  & Xcode 6 support. I'm not sure how people feel about that in general. Here, we generally target the last two OS versions for the most part.

phatmann:
One of KIF's claims to fame is its iOS 6 support. That said, I am ready to drop it if we have come to an impasse. With iOS 10 looming, this might be the time. Thoughts everyone?

mikelupo:
Can we do it in a parallel branch for a cooling off period?

jmkk:
I would love to investigate and address the issue, but Xcode 7.1 simulators are not functional on El Capitan, and I do not have access to a Yosemite machine to debug. 

I'm all for dropping iOS7 support. We could keep a branch for legacy use cases, and move forward on master?

RoyalPineapple:
👍 to a legacy branch and moving forward.

phatmann:
I have opened a pull request for dropping iOS 7 support: #885. I have started by removing Travis support for iOS 7 and updating the project/pod settings to make iOS 8.4 the minimum OS. Please add any other changes needed/desired there.

PR #862, opened 2016-05-25
Added textFromView: function to assert on expected text

jmkk:
Utility function which works with labels, text views and text fields.
Allows to assert on expected text after typing in.


RoyalPineapple:
I'm feeling a little confused as to the use case here, most of the text entry methods already assert that the text is set properly, while the `expectedResult:` allows customization for validating text formatters and the like. What additional benefit does this method this provide?

jmkk:
This is not for validating text entry methods. There are cases where other user actions cause text in text fields/views/labels etc. to get changed. A common case might be a label which gets updated (say, count of likes after user has typed the like button), and we want to assert on the new value. Or imagine testing a feature where user types in some text in a text view, and then can press a button that formats the entered text in a certain way. 

Of course, the same can be achieved by just getting the view by identifier/label, cast it to the appropriate type, and get the text for assertions. It is a convenience function that does it for you.

PR #863, opened 2016-05-25
Added "setText:intoViewWithAccessibilityIdentifier:" helper function

jmkk:
There are cases where text should be entered without typing simulation.
For example, cases where text is pasted, or where we do repeated text field inputs and want to have more performant tests.
This API sets the text explicitly on the view, without going through the Typist.


phatmann:
Why are we only adding this to Identifier tests?

jmkk:
Let me clean up, squash and push again. Thanks.

PR #869, opened 2016-06-01
Accessibility Frame Fallback

fritzsche91:
Currently, if an test **swipes/scrolls** a view, KIF first waits for an accessibility element (UIAccessibilityElement) and the according view, which should the be swiped/scrolled afterwards.

However, in my case (older version of KIF) ```waitForAccessibilityElement:view:withIdentifier:tappable:``` returned a **valid view**, but the **accessibility element** was still **nil** afterwards. Therefore ```swipeAccessibilityElement:inView:inDirection:``` always used a start point with origin:(0,0), width:0, height:0 (accessibilityFrame of accessibility element - nil).

This **zero start point** lead to failing tests, because side navigation bar got swiped (from screen border) instead of the specified view (center of view). As the documentation says *"... and swipe the screen in the given direction from the view's center."*, I personally would prefer that there is always a **fallback frame**, which could be used to get the **center of the view**.

As I have realized, the current ```waitForAccessibilityElement:view:withIdentifier:tappable:``` looks different than the one I am using. However, I still want to share the accessibility frame fallback solution, which solved my issues, because this **centralized fallback approach** should still be an improvement as it potentially increases **stability & maintainability**.
___
**Additional**: This fallback solution was **already used** at other places. These changes only centralise it by using a **separate method**. Additionally, ```tappablePointInElement:andView:``` gets introduced, which allows us to get a tappable point from an accessibility element.

**Attention**: This pull request might *break existing KIF tests*, because it changes some core functionality (tapping, scrolling, swiping) in the case an accessibility frame is not set.

### DIFF
--- a/Classes/KIFUITestActor.m
+++ b/Classes/KIFUITestActor.m
@@ -220,16 +220,8 @@
     [self runBlock:^KIFTestStepResult(NSError **error) {
         
         KIFTestWaitCondition(view.isUserInteractionActuallyEnabled, error, @"View is not enabled for interaction: %@", view);
-        
-        // If the accessibilityFrame is not set, fallback to the view frame.
-        CGRect elementFrame;
-        if (CGRectEqualToRect(CGRectZero, element.accessibilityFrame)) {
-            elementFrame.origin = CGPointZero;
-            elementFrame.size = view.frame.size;
-        } else {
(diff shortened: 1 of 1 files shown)


fritzsche91:
Potentially **related to**:
https://github.com/kif-framework/KIF/pull/839

phatmann:
This is a solid change. I like removing dup code. Since all KIF unit tests pass I feel good about the impact on existing clients. Merging now.

PR #871, opened 2016-06-06
Animation timeout improvements

justinseanmartin:
Animation timeout improvements:
 * Expose out a `waitForAnimationsToFinish` method variant to control how long we sleep before starting to look for animations
 * Rather than taking the existence of an animation key as being sufficient knowledge of if it is complete, instead calculate when the animation should be completed and compare that against the current time.

### DIFF
--- a/Additions/CAAnimation+KIFAdditions.m
+++ b/Additions/CAAnimation+KIFAdditions.m
@@ -1,0 +1,26 @@
+//
+//  CAAnimation+KIFAdditions.m
+//  Pods
+//
+//  Created by Justin Martin on 6/6/16.
+//
+
+#import "CAAnimation+KIFAdditions.h"
+
+
+@implementation CAAnimation (KIFAdditions)
+
+- (double)KIF_completionTime;
+{
+    if (self.repeatDuration > 0) {
+        return self.beginTime + self.repeatDuration;
+    } else if (self.repeatCount == HUGE_VALF) {
+        return HUGE_VALF;
+    } else if (self.repeatCount > 0) {
+        return self.beginTime + (self.repeatCount * self.duration);
+    } else {
+        return self.beginTime + self.duration;
+    }
+}
+
+@end
--- a/Additions/CAAnimation+KIFAdditions.h
+++ b/Additions/CAAnimation+KIFAdditions.h
@@ -1,0 +1,15 @@
+//
+//  CAAnimation+KIFAdditions.h
+//  Pods
+//
+//  Created by Justin Martin on 6/6/16.
+//
+
+#import <QuartzCore/QuartzCore.h>
+
+
+@interface CAAnimation (KIFAdditions)
+
+- (double)KIF_completionTime;
+
+@end
--- a/Additions/CALayer-KIFAdditions.m
+++ b/Additions/CALayer-KIFAdditions.m
@@ -9,8 +9,19 @@
 //
 
 #import "CALayer-KIFAdditions.h"
+#import "CAAnimation+KIFAdditions.h"
+
 
 @implementation CALayer (KIFAdditions)
+
+- (float)KIF_absoluteSpeed
+{
+    __block float speed = 1.0f;
+    [self performBlockOnAncestorLayers:^(CALayer *layer) {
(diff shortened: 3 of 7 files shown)


Additions/CAAnimation+KIFAdditions.m
```diff
@@ -0,0 +1,26 @@
...
+@implementation CAAnimation (KIFAdditions)
+
+- (double)KIF_completionTime;
+{
+    if (self.repeatDuration > 0) {
+        return self.beginTime + self.repeatDuration;
+    } else if (self.repeatCount == HUGE_VALF) {
+        return HUGE_VALF;
+    } else if (self.repeatCount > 0) {
+        return self.beginTime + self.repeatCount * self.duration;
```
RoyalPineapple: Can we add some parenthesis here to clarify the order of operations? Just for readability

RoyalPineapple:
+1 this looks great, can definitely improve the reliability and speed of our tests.

justinseanmartin:
Whoops, don't merge yet, found a mistake.

justinseanmartin:
@kif-framework/collaborators - Any thoughts on this one?

justinseanmartin:
Made one other small improvement here. Layers can have future scheduled animations, which I wasn't accounting for. I'm now making sure that any animations are currently running.

I'm testing this locally by running all of our app tests with a 15 second stabilization timeout and putting a breakpoint on this line:
https://github.com/kif-framework/KIF/pull/871/files#diff-bfb168d5954b87f88ef57cb2ca64bd30R201

With these changes, I'm no longer seeing this breakpoint get hit all the time, which means to me that we're properly detecting the existence and completion of running animations.

justinseanmartin:
@phatmann @bnickel @mikelupo - Any concerns with this change? With the default `waitForAnimationsToFinish` timings, this code path doesn't get exercised, so it only impacts projects that are specifically trying to optimize their wait times.

Specifically, we'd like to use this to avoid some race conditions in our code where we're trying to interact while animations are still running, but we also don't want to add half second sleeps all over our code.

mikelupo:
I'm new enough to KIF that I have to defer to @bnickel and @phatmann. I too want to give this code a try in the future. Probably in our next development lull.

phatmann:
Merging now. After this PR has proven itself, should we make this the default method of waiting for animations?

mikelupo:
@phatmann I'd be ok with that. 
So long as the old way is deprecated and not immediately removed.

Thanks for merging

justinseanmartin:
Thanks @phatmann & @RoyalPineapple for the reviews.

I know that one point the default behavior was changed to wait a maximum of 3 seconds by default:

```
commit e26b3e869d40b9a2a3518ddab05e5cd4cc0d546b
Author: Hendrik von Prince <hendrik.von.prince@sipeso.com>
Date:   Mon Nov 10 13:56:55 2014 +0100

    adding the method waitForAnimationsToFinish to KIFTestActor: It tries to guess if there are any unfinished animations and waits for a certain amount of time (3 seconds) to let them finish. This improves the stability of our integration-tests among the different iPad-Generations.
```

This was generally seen as undesirable, because more often than not it was just sitting for 3 seconds for no reason. I now understand that this was because it wasn't being smart about if an animation was actually running or not. I believe my changes help with this situation. 

I think in our project, I'll lower the animation stabilization timeout and increase the max wait timeout by default. If things go well, I'll report back. It would be interesting for others to try the same in their projects to see if this introduces reliability issues or just makes tests run faster.

Having reduced wait timers with `relativeToAnimationSpeed` has brought some additional flakiness in our tests, so I'm working on fixing those internally now.

justinseanmartin:
@phatmann - Any chance you'd mind doing a dot release with these changes?

phatmann:
Let's wait a day to see if we can get resolution on #859 and #863. Then I will do a new release in any case. Thanks for the request.

phatmann:
A few more things need to settle down before we release. Sure would be nice to have a green Travis before I do :-)

phatmann:
I want to resolve (or triage) a few of the regressions before we release.

PR #872, opened 2016-06-06
Add some additional debugging on failure

justinseanmartin:
Allow opting into having the full view hierarchy dumped on failure. Seeing some discrepancies between my local environment and our CI system, and thought this would help to figure out what's going on.


RoyalPineapple:
This looks great. Awesome addition.

PR #875, opened 2016-06-15
Added "setText:..." helper functions

jmkk:
There are cases where text should be entered without typing simulation.
For example, cases where text is pasted, or where we do repeated text field inputs and want to have more performant tests.
This API sets the text explicitly on the view, without going through the Typist.

Note- this replaces https://github.com/kif-framework/KIF/pull/863

### DIFF
--- a/Classes/KIFUIViewTestActor.m
+++ b/Classes/KIFUIViewTestActor.m
@@ -257,6 +257,14 @@
     [self.actor enterTextIntoCurrentFirstResponder:text fallbackView:fallbackView];
 }
 
+- (void)setText:(NSString *)text;
+{
+    KIFUIObject *found = [self _predicateSearchWithRequiresMatch:YES mustBeTappable:NO];
+    if ([found.view respondsToSelector:@selector(setText:)]) {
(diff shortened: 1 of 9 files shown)


phatmann:
Looking good. Can you also add the corresponding methods to UIViewTestActor?

jmkk:
Done @phatmann

phatmann:
It all looks good. One little picky change, noted as a line comment. Then I can merge.

Classes/KIFUIViewTestActor.m
```diff
@@ -230,6 +230,14 @@ - (void)enterText:(NSString *)text;
     [self enterText:text expectedResult:nil];
 }
 
+- (void)setText:(NSString *)text;
```
phatmann: Can you move this method below the next method so the `enterText` methods stay together?

jmkk:
@phatmann moved.

phatmann:
@jmkk At the risk of trying your patience, can you please update all the files so that setText is not in between the enterText method variations? Sorry I did not notice this before :-(

jmkk:
Of course.

Issue #876, opened 2016-06-20
Support for iOS 10


steipete:
KIF currently does not work with iOS 10b1. Let's use this issue to track the status.

I haven't spent much time on it yet but the first problem here is that Apple removed the old accessibility inspector. These calls fail because the methods no longer exist:

```
self.initialAccessibilityInspectorSetting = [self.axSettingPrefController AXInspectorEnabled:nil]; [self.axSettingPrefController setAXInspectorEnabled:@(YES) specifier:nil];
```

There's no replacement because the whole inspector is gone - you can call `_shortMethodDecription` on the `AccessibilitySettingsController` object to see for yourself. The setting is also gone. The inspector is now a Mac app:

<img width="1126" alt="screen shot 2016-06-20 at 08 11 21" src="https://cloud.githubusercontent.com/assets/58493/16184770/46b5a01c-36bf-11e6-88cf-21adff88b9c9.png">

I'm not completely familiar with the codebase so we need to see how important the inspector is and if we can invoke the same touches without it. Did anyone already work on compatibility with Xcode 8/iOS 10b1?

phatmann:
With #877 merged, do you still want to leave this PR open?

steipete:
iOS 10 works now.

phatmann:
Woot! Thanks for your help here.

PR #877, opened 2016-06-20
iOS 10: Use alternate way to enable accessibility

steipete:
This addresses #876 based on what Google is doing in EarlGrey.

https://github.com/google/EarlGrey/pull/148/files

We now use this branch internally at PSPDFKit and Jenkins is green fron 8.4 to 10.0b1. (well, iOS 10 has some troubles but they are not KIF related...)

We no longer show the accessibility inspector, but I actually feel that's a good thing, and it's gone in iOS 10 anyway. The old logic has to remain until KIF drops support for iOS 8.

### DIFF
--- a/Classes/KIFAccessibilityEnabler.m
+++ b/Classes/KIFAccessibilityEnabler.m
@@ -46,12 +46,12 @@
 {
     // This works as of iOS 9.
     CFPreferencesSetAppValue((CFStringRef)@"ApplicationAccessibilityEnabled",
-                             enabled, (CFStringRef)@"com.apple.Accessibility");
+                             (__bridge CFPropertyListRef)(@(enabled)), (CFStringRef)@"com.apple.Accessibility");
     CFPreferencesSynchronize((CFStringRef)@"com.apple.Accessibility",
                              kCFPreferencesCurrentUser, kCFPreferencesAnyHost);
     CFNotificationCenterPostNotification(CFNotificationCenterGetDarwinNotifyCenter(),
                                          (CFStringRef)@"com.apple.accessibility.cache.app.ax",
-                                         nil, nil, kCFBooleanTrue);
+                                         nil, nil, YES);
 }
 
 - (void)enableAccessibility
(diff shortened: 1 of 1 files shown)


steipete:
Not sure why tests don't run. Things work great locally and on our Jenkins.

Classes/KIFAccessibilityEnabler.m
```diff
@@ -38,8 +42,39 @@ + (instancetype)sharedAccessibilityEnabler
...
+- (void)setApplicationAccessibilityEnabled:(BOOL)enabled
+{
+    // This works as of iOS 9.
+    CFPreferencesSetAppValue((CFStringRef)@"ApplicationAccessibilityEnabled",
+                             kCFBooleanTrue, (CFStringRef)@"com.apple.Accessibility");
+    CFPreferencesSynchronize((CFStringRef)@"com.apple.Accessibility",
+                             kCFPreferencesCurrentUser, kCFPreferencesAnyHost);
+    CFNotificationCenterPostNotification(CFNotificationCenterGetDarwinNotifyCenter(),
+                                         (CFStringRef)@"com.apple.accessibility.cache.app.ax",
+                                         nil, nil, enabled);
```
axi0mX: The last argument for CFNotificationCenterPostNotification is "deliverImmediately":

"If true, the notification is delivered to all observers immediately, even if some observers are in suspended (background) applications and they requested different suspension behavior when registering for the notification. If false, each observer’s requested suspension behavior is respected.

If center is a Darwin notification center, this value is ignored."

While the value is ignored, and thus should not affect execution, it would be best to just set it to TRUE (instead of YES, since it is type Boolean and not BOOL, or variable enabled, since there is no change in behavior either way) to make the code easier to read.

steipete: Good point, fixed!

phatmann:
Change looks good, as far as I can tell. Which is not very far ;-)

Let's merge it, since it preserves backward compatibility. We will need the community input on whether it works for everyone.

phatmann:
This was reverted via #883 since it caused issues in Xcode 7.

PR #878, opened 2016-06-22
Call willDisplay on tableview cells

justinseanmartin:
We have a situation where we have some elements haven't yet been fully initialized and thus don't get the proper accessibilityLabel until after the `willDisplayCell` UITableViewDelegate method is called. 

It seemed reasonable to add this here, though we could alternatively change the app to set this information on creation of the cell as opposed to just before it is being displayed.

Thoughts?


justinseanmartin:
I'm going to self-reject and just change the behavior in our app.

mikelupo:
Would the lack of this potentially be the cause of a crash or exception?

PR #879, opened 2016-06-23
Don't change the typing delay based on animation speed

justinseanmartin:
I've found in our project when attempting to bump to the latest KIF that sometimes attempting to tap a button directly after typing to be randomly unreliable (tap doesn't actually get invoked). I'm having a hard time reproducing this locally, but it happens on our CI machines (dual core mac minis) often enough to be nearly unusable (every 2-3 runs has a failure).

Rather than using `KIFRunLoopRunInModeRelativeToAnimationSpeed` for modifying the typing delay, I believe test authors should explicitly set `KIFTypist.keystrokeDelay`. The `keyWindow`'s layer animation speed has nothing to do with the keyboard window animations.  However, lowering this value hits the reliability issues I mentioned before. Rather than waiting the .01s between every tap, instead only wait the full .01s after the last keypress (which seemed to always work fine for us before).

The timings we're currently using in our base test class `-beforeEach`:

```
    UIApplication.sharedApplication.animationSpeed = 4.0;
    KIFTypist.keystrokeDelay = 0.0025f;
    KIFTestActor.defaultAnimationStabilizationTimeout = 0.1;
    KIFTestActor.defaultAnimationWaitingTimeout = 2.0;
```

I also fixed the indentation of the method in `KIFUITestActor.m`, so it reviews much better with `&w=1` on the github URL.

### DIFF
--- a/Classes/KIFTypist.h
+++ b/Classes/KIFTypist.h
@@ -13,7 +13,10 @@
 + (void)registerForNotifications;
 + (BOOL)keyboardHidden;
 + (BOOL)enterCharacter:(NSString *)characterString;
+
++ (NSTimeInterval)keystrokeDelay;
 + (void)setKeystrokeDelay:(NSTimeInterval)delay;
+
 + (BOOL)hasHardwareKeyboard;
 + (BOOL)hasKeyInputResponder;
 
--- a/Classes/KIFTypist.m
+++ b/Classes/KIFTypist.m
@@ -89,8 +89,13 @@ static NSTimeInterval keystrokeDelay = 0.01f;
         [[UIKeyboardImpl sharedInstance] addInputString:characterString];
     }
     
-    KIFRunLoopRunInModeRelativeToAnimationSpeed(kCFRunLoopDefaultMode, keystrokeDelay, false);
+    CFRunLoopRunInMode(kCFRunLoopDefaultMode, keystrokeDelay, false);
     return YES;
+}
+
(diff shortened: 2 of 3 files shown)


Classes/KIFTypist.h
```diff
@@ -13,7 +13,10 @@
 + (void)registerForNotifications;
 + (BOOL)keyboardHidden;
 + (BOOL)enterCharacter:(NSString *)characterString;
+
++ (NSTimeInterval)keystrokeDelay;
```
RoyalPineapple: Why don't we just make this a regular `@property` we aren't doing anything in the setter/getter

RoyalPineapple:
This looks good. I'd like to see travis go green but it looks like we've got some flakey tests anyhow.

justinseanmartin:
Tests appear to be generally flakey on latest master:
https://travis-ci.org/kif-framework/KIF/builds

Seems to have been introduced by this PR getting merged:
https://github.com/kif-framework/KIF/pull/852

Classes/KIFTypist.h
justinseanmartin: It is a static on the class, rather than an instance method.

RoyalPineapple: Of course. thanks.

justinseanmartin:
@phatmann - Thoughts on merging this?

phatmann:
Looks good, I like the approach.

PR #880, opened 2016-06-23
Revert "Expose logic for matching accessibility elements, and use it …

justinseanmartin:
…to fix bugs with cell reuse"

This reverts commit 4575a4b00eec713a294bfd3c7f2d1967be18653c.

We don't necessarily need to merge this, but I'd like to at least see if this makes Travis happier. If so, we can decide if we want to merge this, if we want to take time to track down why this is breaking integration tests, or if we want to temporarily disable the flakey tests.

### DIFF
--- a/Additions/UIAccessibilityElement-KIFAdditions.m
+++ b/Additions/UIAccessibilityElement-KIFAdditions.m
@@ -55,13 +55,6 @@ MAKE_CATEGORIES_LOADABLE(UIAccessibilityElement_KIFAdditions)
         return NO;
     }
     
-    // viewContainingAccessibilityElement:.. can cause scrolling, which can cause cell reuse.
-    // If this happens, the element we kept a reference to might have been reconfigured, and a
-    // different element might be the one that matches.
-    if (![UIView accessibilityElement:element hasLabel:label accessibilityValue:value traits:traits]) {
-        return NO;
-    }
-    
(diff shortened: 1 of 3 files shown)


justinseanmartin:
Ok, so some of the failures go away, but there is still some flakey failures. I just kicked it off again to see if it ever will go all green.

justinseanmartin:
3 different shards flaked out for 3 different reasons. Maybe we should leave the changes in, and work on instead trying to stabilize the test suite overall.

justinseanmartin:
I kicked it one more time to see.

justinseanmartin:
It looks like #861 also introduced another regression that only impacts the iPad 2 iOS 7 configuration, which is the other consistent failure I'm seeing.

phatmann:
I will wait a little bit to see what @jmmk finds in #861.

justinseanmartin:
I think possibly #877 also broke CI for Xcode 7 (just picked that up in a rebase). We should probably add an additional Travis CI config for Xcode 8 as well.

phatmann:
Rerunning tests. Note that we might need to revert #877 to see where we stand.

phatmann:
#877 has been reverted via #883. We will see if this helps out with Travis.

justinseanmartin:
Been open for a really long time, and haven't needed it, so closing. Travis is green at the moment.

PR #896, opened 2016-08-16
Fix passing nil into a method that has an alternative with 1 fewer arguments

voznesenskym:
No reason to pass nil in. In fact, for nullability we get an error w/ this at Uber.


voznesenskym:
Thanks @mikelupo

PR #899, opened 2016-08-24
Fix for really slow stack overflow when tableview is removed from view

justinseanmartin:
We ran into a recent test failure within KIF that was fun and convoluted:

1. We're looking for an element on the screen to tap
2. We find a UITableView that is in a UINavigationTransitionView (animating it into place)
3. We find the UIView we want to tap inside the UITableView, but it isn't tappable at its current location, so we try to scroll it into view
4. After scrolling, we sleep for a half second and check to see if it is now tappable
5. After that half second, the UINavigationTransitionView is no longer part of the window (because the animation transition completed)
6. From this point on, we'll recursively attempt to scroll the element into view, wait half a second, and then check to see if it is visible again

It sits recursing infinitely in this loop with a half second sleep between each call (so it doesn't quickly run out of stack space) and then eventually blows up.

My solution was to bail out if we find that we're checking an element that is no longer a part of the view hierarchy.

### DIFF
--- a/Additions/UIView-KIFAdditions.m
+++ b/Additions/UIView-KIFAdditions.m
@@ -223,7 +223,7 @@ NS_INLINE BOOL StringsMatchExceptLineBreaks(NSString *expected, NSString *actual
         }
     }
     
-    if (!matchingButOccludedElement) {
+    if (!matchingButOccludedElement && self.window) {
         if ([self isKindOfClass:[UITableView class]]) {
             UITableView *tableView = (UITableView *)self;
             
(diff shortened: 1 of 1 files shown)


justinseanmartin:
This actually doesn't work as I'd expected. I need to do another iteration on this before it can be merged.

justinseanmartin:
Ok, I take it back. It does work as expected, I was just seeing a failure caused by another reason in a similar test in our suite. I think this should work fine as it is.

phatmann:
Travis tests all passed, but build is marked as failing due to this [error](https://github.com/travis-ci/travis-ci/issues/6522). Hopefully Travis folks will fix it soon.

PR #901, opened 2016-08-30
Modify CI config to cover 8.1, modify podspec

### LINKED ISSUE #900: Why the hard cutoff for support at 8.4?
My app currently supports iOS 8.0 and up and was curious why the hard cutoff at 8.4 vs 8.0? I can't find anything in the commits that prevents this. I wasn't sure how to get this question answered so I created an issue.

justinseanmartin:
Resolves #900


justinseanmartin:
Looks like Xcode 7.2 doesn't have the 8.1 simulators installed. Checking Xcode 7.3 now. If that doesn't work, I'll try dropping it to Xcode 6.4.

justinseanmartin:
Xcode 7.3 image didn't contain it either, changing the Xcode 8.4 config to run on an 8.1 simulator instead.

justinseanmartin:
I removed the explicit settings on the targets so that they all inherit the deployment target from the project instead. I'm thinking that we shouldn't use the dynamic framework CI target for this though, as that is really just making sure that we can build and link properly. Gonna change the iPad Air config to run on Xcode 6.4 w/ iOS 8.1.

phatmann:
Thanks, @justinseanmartin. All green, baby!

PR #916, opened 2016-09-29
Include element and view in 'is not tappable' error details

erikprice:
Minor logging improvement that has been helping us troubleshoot errors.


phatmann:
Nice little change, thanks!

PR #939, opened 2017-01-23
Convert paths to be in window coordinates before we start moving the window

thorthugnasty:
Converts coordinates from view coordinates to be relative to the window.


justinseanmartin:
Just tried these changes against our full test suite and ran into some failures. I'm taking a look now to see what is going wrong.

justinseanmartin:
Verified that things work with the latest changes on our tests. Change makes sense to me, but would be nice to have a second set of eyes.

phatmann:
Looks good. Maybe this will solve some of the mysterious problems this method has shown over the years.

PR #940, opened 2017-01-23
Don't enumerate hidden windows

thorthugnasty:
When we find a hidden window nothing should be visible beyond that view.


phatmann:
Good idea.

PR #949, opened 2017-02-26
Support for selection in DatePickers with Minimum Date value

zaxy78:
Currently,
when trying to use `selectDatePickerValue` on DatePickers with minimumDate set, selection fails with no error. 

After debugging I realized - since Apple push into the DatePickers a 1000 rows of data, when setting minimum date - for example Today - the default value will be in the middle (row 500).

Now when the `selectDatePickerValue` method is looking for a future date (e.g "Feb 28") it will start in row 0 and find "Feb 28" of 2016. upon selecting that value, the DatePicker View will jump back automatically to minimum value (Today) and will not select the desired date.

This pull-request suggest a fix. 
I am open for changes, feedback, etc.
I tried the most simple approach i could think of:
1) if this is a DatePicker & Minimum Date is not nil 
2) find Minimum value row (by selecting it) and start your search loop for that row forward.

Would love to see it pushed in :)

### DIFF
--- a/Classes/KIFUITestActor.m
+++ b/Classes/KIFUITestActor.m
@@ -622,14 +622,30 @@
         }
         // Find the picker view
         UIPickerView *pickerView = nil;
+        NSInteger firstRowToSearchIndex = 0;
         switch (pickerType)
         {
             case KIFUIDatePicker:
+            {
                 pickerView = [[[[UIApplication sharedApplication] datePickerWindow] subviewsWithClassNameOrSuperClassNamePrefix:@"UIPickerView"] lastObject];
                 KIFTestCondition(pickerView, error, @"No picker view is present");
+                
+                // If minimum date limit exist, search the value from minimum and beyond.
+                UIDatePicker *datePicker = (UIDatePicker *)pickerView;
+                if (datePicker.datePickerMode == UIDatePickerModeDateAndTime
+                    && datePicker.minimumDate != nil) {
+                    
+                    [datePicker setDate:datePicker.minimumDate];
+                    firstRowToSearchIndex = [pickerView selectedRowInComponent:0];
+                }
+                
+                
                 break;
+            }
             case KIFUIPickerView:
+            {
                 pickerView = [[[[UIApplication sharedApplication] pickerViewWindow] subviewsWithClassNameOrSuperClassNamePrefix:@"UIPickerView"] lastObject];
+            }
         }
         
         NSInteger componentCount = [pickerView.dataSource numberOfComponentsInPickerView:pickerView];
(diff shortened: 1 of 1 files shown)


mikelupo:
Done.
Thank you for your contribution to KIF!

justinseanmartin:
For these types of changes, it would be great to include a new functional test in the Test Host that shows the problematic behavior that you're encountering and shows that it is working correctly.

Classes/KIFUITestActor.m
```diff
@@ -622,22 +622,38 @@ - (void)selectPickerValue:(NSArray *)pickerColumnValues pickerType:(KIFPickerTyp
...
+            {
                 pickerView = [[[[UIApplication sharedApplication] datePickerWindow] subviewsWithClassNameOrSuperClassNamePrefix:@"UIPickerView"] lastObject];
                 KIFTestCondition(pickerView, error, @"No picker view is present");
+                
+                // If minimum date limit exist, search the value from minimum and beyond.
+                UIDatePicker *datePicker = (UIDatePicker *)pickerView;
+                if (datePicker.datePickerMode == UIDatePickerModeDateAndTime
+                    && datePicker.minimumDate != nil) {
+                    
+                    [datePicker setDate:datePicker.minimumDate];
```
justinseanmartin: Ok, so I took a closer look and now understand the problem a bit better. The problem with `UIDatePickerModeDateAndTime` is that it has multiple years worth of values matching `Feb 28`, but it isn't explicitly called out anywhere in the field's accessibility.

Besides the bug that was called out regarding `firstRowToSearchIndex`, I think the proposed solution works, but it only fixes a particular use case. 

One alternative solution might be to add a options enum `KIFPickerSearchOrder` like:

```
typedef NS_OPTIONS(NSUInteger, KIFPickerSearchOrder) {
    KIFPickerSearchForwardFromStart = 0,
    KIFPickerSearchBackwardFromEnd = 1,
    KIFPickerSearchForwardFromCurrentValue = 2,
    KIFPickerSearchBackwardFromCurrentValue = 3
};
```

I think this would achieve the same results as what you've currently got, except the behavior is less surprising to callers.

My favored approach is to add a new API for `-[KIFUITestActor selectDatePickerDate:(NSDate *)date]` that stores the selected rows for each component, calls through to `-[NSDatePicker setDate:animated:]`, spins the runloop, then calls the delegate for `pickerView:didSelectRow:inComponent:` for each component whose row changed. I think this would be the most straightforward approach. Given that we're already using `selectRow:` picker view API and directly invoking the delegate rather than using flick gestures to scroll the picker view, this feels ok to me. Thoughts?

Classes/KIFUITestActor.m
```diff
@@ -622,22 +622,38 @@ - (void)selectPickerValue:(NSArray *)pickerColumnValues pickerType:(KIFPickerTyp
...
+            }
         }
         
         NSInteger componentCount = [pickerView.dataSource numberOfComponentsInPickerView:pickerView];
         KIFTestCondition(componentCount == columnCount, error, @"The UIDatePicker does not have the expected column count.");
         
         for (NSInteger componentIndex = 0; componentIndex < componentCount; componentIndex++) {
             NSInteger rowCount = [pickerView.dataSource pickerView:pickerView numberOfRowsInComponent:componentIndex];
-            for (NSInteger rowIndex = 0; rowIndex < rowCount; rowIndex++) {
+            for (NSInteger rowIndex = firstRowToSearchIndex; rowIndex < rowCount; rowIndex++) {
```
justinseanmartin: Ah, nevermind, there are 10k rows for each component, not just for the first one. That is pretty bizarrely implemented by Apple. I guess this works as is.

Classes/KIFUITestActor.m
zaxy78: Sounds great. 
I was tring to think of a more general solution than my fix. 
your suggestion looks great. Gives more control back to the caller.

I will implement it and push. should I open another pull request or can I use this one?

justinseanmartin: New PR would be great, leaving this one merged.

PR #951, opened 2017-03-02
Added KIFUITestActor-IdentifierTests to KIF-Header-File

LordNali:
Added KIFUITestActor-IdentifierTests to KIF-Header-File

### DIFF
--- a/Classes/KIF.h
+++ b/Classes/KIF.h
@@ -12,6 +12,7 @@
 #import "KIFSystemTestActor.h"
 #import "KIFUITestActor.h"
 #import "KIFUITestActor-ConditionalTests.h"
+#import "KIFUITestActor-IdentifierTests.h"
 
 #import "KIFUIViewTestActor.h"
 #import "KIFUIObject.h"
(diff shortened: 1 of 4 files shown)


phatmann:
This header is purposely not part of KIF.h, because it is brought in via a CocoaPod subspec. Can you explain what is motivating you to make this change?

LordNali:
Okay, I didn't know that - we tried to use Carthage to integrate KIF in our Project (like for all our other Frameworks), but unfortunately that doesn't work as expected. Then we included the KIF Project manually / from Source and that works fine.

I'm a bit confused, since your ReadMe list 'Installation (from GitHub)' as an Install-Option - are we're the only ones using this option with Swift and AccessibilityIdentifier Tests?!

phatmann:
Can't you just include the Identifier test header manually alongside KIF.h?

LordNali:
Sure - I just added the Identifier Test header to our Objective-C bridging header. So no need to merge this PR, but maybe this should be mentioned in the ReadMe.

phatmann:
Sounds good. If you feel like updating the README please go for it, otherwise I can do so. Thanks.

LordNali:
I would kindly ask you to update the ReadMe - I'm not sure my English fits the purpose here. Furthermore I would recommend that you implicitly make a Section for "Swift-Users" and point them directly to the Testable-Swift Example: In our case it was easy to configure our Test Target by comparing our Project with your Example Project side by side.

phatmann:
Done!

PR #952, opened 2017-03-02
Add date picker search order

zaxy78:
After discussion with @justinseanmartin on my last pull-request #949 

Here is the suggested wider fix with _KIFPickerSearchOrder_

```
typedef NS_ENUM(NSUInteger, KIFPickerSearchOrder) {
    KIFPickerSearchBackwardFromEnd = -2,
    KIFPickerSearchBackwardFromCurrentValue = -1,
    KIFPickerSearchForwardFromCurrentValue = 1,
    KIFPickerSearchForwardFromStart = 2,
};

- (void)selectDatePickerValue:(NSArray *)datePickerColumnValues withSearchOrder:(KIFPickerSearchOrder)searchOrder

- (void)selectPickerViewRowWithTitle:(NSString *)title inComponent:(NSInteger)component withSearchOrder:(KIFPickerSearchOrder)searchOrder

```

I had one small problem with AM:PM, and I fixed it with another IF for that specific case. 
Any other creative solutions will be gladly received.

### DIFF
--- a/Classes/KIFUITestActor.h
+++ b/Classes/KIFUITestActor.h
@@ -48,10 +48,10 @@ typedef NS_ENUM(NSUInteger, KIFPickerType) {
  @constant KIFPickerSearchBackwardFromCurrentValue Search from current value backwards.
  */
 typedef NS_ENUM(NSUInteger, KIFPickerSearchOrder) {
-    KIFPickerSearchBackwardFromEnd = -2,
-    KIFPickerSearchBackwardFromCurrentValue = -1,
-    KIFPickerSearchForwardFromCurrentValue = 1,
-    KIFPickerSearchForwardFromStart = 2,
+    KIFPickerSearchForwardFromStart = 0,
+    KIFPickerSearchBackwardFromEnd = 1,
+    KIFPickerSearchForwardFromCurrentValue = 2,
+    KIFPickerSearchBackwardFromCurrentValue = 3
 };
 
 /*!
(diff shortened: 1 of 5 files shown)


phatmann:
All looks good except for that AM/PM hack. Can you explain what the issue is and how you are solving it?

zaxy78:
Yap,
The thing is that although in Apple's implementation we have 1000 rows of vaules in the DatePicker columns, the AM:PM is only 2 rows. (of course). 
What happens there is that there is no point to **StaftFromCurrent** Forawrd or Backwards. Since if the desired time to pick is 6am, but the current time is 3PM and search order is forward, it will **fail** since you can't go any further after PM, and vice versa. 

So the search order works for the cyclic columns with 1000 rows of values, but not for a 2 rows column. So in this case - I simply go back to defaalt order (0 to end, progress forward)

phatmann:
Is there a better test for AM/PM component than checking if row count is 2?

zaxy78:
I can't think of a better test, but I assure you this -

If it's a 2 row component, I'd say scan it regularly anyhow... Since chances are there are no duplicates and you'll find what you need with no problem....

justinseanmartin:
Thanks for making these changes, just had time to review and they look great to me!

justinseanmartin:
Touching the storyboard in Xcode 8 broke compiling the tests on Xcode 7 (causing master CI to be red). Any chance you'd be willing to fix @zaxy78?

@phatmann - We should wait for Travis to go green before merging if we can help it.

phatmann:
@justinseanmartin I already merged this PR, sorry.

PR #954, opened 2017-03-09
Trigger travis

zaxy78:
Trtying to trigger travis again. wondering what failed last time


zaxy78:
triggered. failed again. Not good :(

justinseanmartin:
Looks like this was caused by the storyboard changes, Apple broke backwards compatibility in Xcode 8 for Interface Builder files for usage with Xcode 7.

zaxy78:
Thanks for fixing it :+1:

PR #955, opened 2017-03-10
Fix scrolling to cells that are under a translucent toolbar (#953)

### LINKED ISSUE #953: Unable to tap table view cells that are under a toolbar
I have a `UITableViewController` subclass in a `UINavigationController` with `toolbarHidden = false`. When I use KIF to tap on a cell that is currently under the toolbar, KIF doesn't scroll and it fails. I assume this is because KIF thinks the cell is already on screen, but is unable to touch it because it is obscured by the toolbar.

hiltonc:
Fixes #953

### DIFF
--- a/Additions/UIAccessibilityElement-KIFAdditions.m
+++ b/Additions/UIAccessibilityElement-KIFAdditions.m
@@ -177,6 +177,11 @@ MAKE_CATEGORIES_LOADABLE(UIAccessibilityElement_KIFAdditions)
             
             if (((UIAccessibilityElement *)view == element) && ![view isKindOfClass:[UITableViewCell class]]) {
                 [scrollView scrollViewToVisible:view animated:YES];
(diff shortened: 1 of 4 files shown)


hiltonc:
Below my fix I see this comment:

```objc
// Only call scrollRectToVisible if the element isn't already visible
// iOS 8 will sometimes incorrectly scroll table views so the element scrolls out of view
```

I suspect that this because the scrolling was done using the `UITableViewWrapperView` (a `UIScrollView` subclass). With my change I think that the commented hack is no longer necessary, but I didn't remove it fearing that doing so would break things.

phatmann:
@hiltonc Thanks for the PR and comments. I am going to merge this now.

If you have the time and/or gumption, please try removing the hack you mentioned above and see if all the tests still pass. If so I suggest we remove the hack via another PR, as the last thing KIF needs is extra, unneeded hacks! Thanks again.

PR #957, opened 2017-03-17
Update build script to have travis pass

wolffan:
Travis problem, run tests with derived data option

### DIFF
--- a/scripts/ci.sh
+++ b/scripts/ci.sh
@@ -21,4 +21,4 @@ if [[ ! ${SIMULATOR} =~ .*OS=7.* ]]; then
   env NSUnbufferedIO=YES xcodebuild test -derivedDataPath=${PWD}/build/KIFFramework -scheme KIFFrameworkConsumerTests -destination "platform=iOS Simulator,${SIMULATOR}" | xcpretty -c
 fi
 
-env NSUnbufferedIO=YES xcodebuild test -scheme KIF -derivedDataPath=${PWD}/build/KIF -destination "platform=iOS Simulator,${SIMULATOR}" | xcpretty -c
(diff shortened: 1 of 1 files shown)


justinseanmartin:
I assumed that Travis used VMs and reset them to the default state between each run to ensure there was no pollution from one run to the next. Are you sure this is necessary? Is there documentation somewhere? 

It would at least be nice to have this for people running the CI script on their local machines I suppose.

Our Travis builds are generally broken at the moment and need to be addressed. I'll take care of it over the weekend.

wolffan:
@justinseanmartin there's a problem with xcodebuild when you run your tests. Which times out and gives that error 65 we are seeing. 
This is not really documented (I found it debugging fastlane issues and since fastlane uses xcodebuild I guess this may be similar issue).

That's why I was experimenting, my experiment is not yet finished :).
But I truly believe the error is related with the build taking too long and then travis failing on the test run.
BTW using something like fastlane for testing is completely out of question?
The idea of this pr was to fix the CI :)

justinseanmartin:
Ah, the Travis CI failure isn't due to a timeout issue, it is due to the error message: `The document "(null)" requires Xcode 8.0 or later.`

This was introduced by #952. The fix is to revert the changes to the storyboard and redo them in Xcode 7. We could alternatively bump the CI config to run on Xcode 8, but that would mean dropping support for Xcode 7, which I'm not sure that we're ready to do yet.

I'm not opposed to changing the CI configuration to run with fastlane, as long as it passes consistently and the logs are easily parseable.

justinseanmartin:
CI should go green with changes in #958

phatmann:
Re-running tests now

wolffan:
As mentioned this wasn't related to the original crash :/.
There's an improvement to be done when running xcodebuild, but i'll create a new one with more detail :)

zaxy78:
@wolffan I was the one to introduce this Travis failure. 
I edited the te MainStoryboard with Xcode8. Didn't know Travis is using Xcode7.
Thanks @justinseanmartin for the fix.

But I have a question to you - do you also have problems with runnin gKIF tests via fastlane?

we use it here and it hangs/get stuck. see my issue #961 ...

PR #971, opened 2017-05-03
Move load back into KIFTestActor class

justinseanmartin:
Partially reverts changes introduced in #819.

We bumped KIF to the latest release for the first time in a while and found that some logic test targets (e.g. no app host) that happen to link KIF into them as a dynamic framework fail to dlopen at load time. I isolated it back to this change, and it appears to fix the issue that we were seeing.

Ideally, we'd have a different set of test dependencies for logic tests and app tests, but the project has manually managed dependencies and this used to work and was simpler for them to manage. As this used to work with v3.4.2, it seemed appropriate to fix this and partially restore the previous behavior.

I'm not clear if the change of moving the +load to a UIApplication was necessary or just seemed more correct. @benasher44?

I'm not 100% clear on why this is failing, but I also don't know exactly how categories are supposed to behave if UIKit isn't available at runtime. This does fix the issue of KIF being linked via a dynamic framework into a logic test though. 

Generally, it also seems safer to not place a `+load` in a UIApplication category, as you can't use a method prefix and thus might stomp another `+load` in another category or in UIKit. Right?

CC: @benasher44 @holmes @phatmann

### DIFF
--- a/Classes/KIFTestActor.m
+++ b/Classes/KIFTestActor.m
@@ -18,6 +18,18 @@
 
 @implementation KIFTestActor
 
++ (void)load
+{
+    @autoreleasepool {
+        if (NSClassFromString(@"UIApplication")) {
+            [UIApplication swizzleRunLoop];
+            NSLog(@"KIFTester loaded");
+        } else {
+            NSLog(@"KIFTester skipping runloop swizzling, no UIApplication class found.");
+        }
+    }
+}
+
 - (instancetype)initWithFile:(NSString *)file line:(NSInteger)line delegate:(id<KIFTestActorDelegate>)delegate
 {
     self = [super init];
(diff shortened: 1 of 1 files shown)


benasher44:
You'd need to dig into the runtime to see whether or not load category methods "stomp" each other or not. I'm not sure why this PR helps anything, and I'd kind of expect the opposite. Using the `load` category method on `UIApplication` guarantees that the `UIApplication` class is loaded by the time this method as called. Moving it somewhere else removes that guarantee, as far as I understand.

justinseanmartin:
UIApplication (& UIKit in general) are provided by the OS, so I don't think that a test dependency could be loaded before UIKit is loaded. I would've presumed that +load isn't different than any other method that is loaded via a category:

https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/CustomizingExistingClasses/CustomizingExistingClasses.html#//apple_ref/doc/uid/TP40011210-CH6-SW4

Does it seem reasonable to land this change?

benasher44:
Hm that's a good point. I think it's reasonable, but it feels arbitrary without some explanation of why this change is needed. Maybe you could clarify:

> …fail to load time.
?

justinseanmartin:
Sounds good, I'll make that message more clear.

benasher44:
Makes a bit more sense with the updated description. Ship it!

justinseanmartin:
The error that we were seeing is that launching the logic test would fail trying to dlopen the dynamic library that had KIF linked into it. I don't have the logs handy at the moment, but essentially the error message contained the message "<library> can't be loaded because it is damaged or missing necessary resources".

PR #972, opened 2017-05-04
Fix Example Projects and run them via CI

### LINKED ISSUE #970: Testable Swift Example Issue
Hey,

I ran one of example(Testable Swift) from downloaded KIF repo and found below error: 

**Assertions: The step timed out after 10.00 seconds: Failed to find accessibility element with the label "Red"
SimpleObjCTest.m:20**

I found that this issue is related to accessibility and tried to debug but couldn't find anything. 
I am using XCode 8.3.2 and device(or simulator) running iOS 10.3.

Thanks in advance.

justinseanmartin:
Fixes #970

Also, updated it to use the examples to use the v3 API that @RoyalPineapple introduced.

CC: @imsrc21 @phatmann

### DIFF
--- a/Documentation/Examples/Calculator/Acceptance Tests/Acceptance Tests-Prefix.pch
+++ b/Documentation/Examples/Calculator/Acceptance Tests/Acceptance Tests-Prefix.pch
@@ -1,9 +1,0 @@
-//
-// Prefix header for all source files of the 'Acceptance Tests' target in the 'Acceptance Tests' project
-//
-
-#ifdef __OBJC__
-    #import <UIKit/UIKit.h>
-    #import <Foundation/Foundation.h>
-    #import <KIF/KIF.h>
-#endif
--- a/Documentation/Examples/Calculator/Acceptance Tests/BasicCalculatorTests.m
+++ b/Documentation/Examples/Calculator/Acceptance Tests/BasicCalculatorTests.m
@@ -7,56 +7,25 @@
 //  See the LICENSE file distributed with this work for the terms under
 //  which Square, Inc. licenses this file to you.
 
-#import <KIF/KIF.h>
+@import KIF;
 
-@interface KIFUITestActor (BasicCalculator)
+#import "KIFUITestActor+BasicCalculator.h"
 
-- (void)enterValue1:(NSString *)value;
-- (void)enterValue2:(NSString *)value;
-- (void)setOperation:(NSString *)operation;
-- (void)enterValue1:(NSString *)value1 value2:(NSString *)value2 operation:(NSString *)operation;
-- (void)waitForResult:(NSString *)result;
-
-@end
-
-@implementation KIFUITestActor (BasicCalculator)
-
-- (void)enterValue1:(NSString *)value
-{
-    [self clearTextFromAndThenEnterText:value intoViewWithAccessibilityLabel:@"First Number"];
-}
-
-- (void)enterValue2:(NSString *)value
-{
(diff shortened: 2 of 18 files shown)


.travis.yml
```diff
@@ -1,8 +1,8 @@
 language: objective-c
 matrix:
   include:
-    - osx_image: xcode6.4
-      env: 'SIMULATOR="name=iPad Air,OS=8.1"'
+    - osx_image: xcode8.3
```
phatmann: Good idea.

Classes/KIFUITestActor.m
```diff
@@ -387,24 +387,32 @@ - (void)enterTextIntoCurrentFirstResponder:(NSString *)text
...
+    UIView *fallback = fallbackView;
+    if (!fallback) {
+        UIResponder *firstResponder = [[[UIApplication sharedApplication] keyWindow] firstResponder];
+
+        if ([firstResponder isKindOfClass:[UIView class]]) {
+            fallback = (UIView *)firstResponder;
+        }
+    }
+
+    id<UITextInputTraits> textInputField = nil;
```
phatmann: Finally a fix for this issue.

phatmann:
I will run Travis again now.

Classes/KIFUITestActor.m
justinseanmartin: Yea, saw it appearing on the integration tests on iOS10 and thought I'd better fix it. I could just change the app to disable autocorrection, but I figured if I'm already in here, I might as well fix it for all text controls.

justinseanmartin:
I need to take a look at my changes to the CI script, for some reason `Acceptance Tests` is failing on Travis.

Classes/KIFUITestActor.m
justinseanmartin: Unfortunately doesn't seem to work: https://travis-ci.org/kif-framework/KIF/jobs/229223229

```
    ✗ testEnteringEmojiCharactersIntoViewWithAccessibilityLabel, The step timed out after 10.00 seconds: Failed to get text " 😓He😤ll👿o" in field; instead, it was " 😓He😤lol👿o"
```

justinseanmartin:
Fixed the script (was boneheaded mistake), but still the iOS10 job is broken for the `testEnteringEmojiCharactersIntoViewWithAccessibilityLabel` test. I'll need to look more closely at this.

justinseanmartin:
Don't merge this, but I'm seeing if I just make the control itself on initialization to be `UITextAutocorrectionTypeNo` if that will even allow the test to pass. I can't reproduce the failure locally, but I'm still on Xcode 8.2.1 instead of 8.3.2. Don't know if that's it or not though, seems like it shouldn't be.

justinseanmartin:
Apparently the changes in 0604fe2 fix the test, but the changes in ed296f1 do not. Very strange... Will do some more digging.

justinseanmartin:
I'm going to split this out into two PRs. One for the original issue in #972 and a separate one to add an iOS10 Xcode 8.3 build configuration (which has to include fixing that typing test).

justinseanmartin:
Ok @phatmann & @RoyalPineapple, please take a look. This now just updates the sample projects to all pass, adds a CI configuration that runs them, and converts them to use the KIFUIViewTestActor's `viewTester` syntax.

mikelupo:
@justinseanmartin thank you for this update. Looks like Travis is happy. I like the change. +1 from me.

PR #974, opened 2017-05-11
Add a CI config for Xcode 8 w/ iOS 10

justinseanmartin:
Add a CI config for Xcode 8 w/ iOS 10. This requires fixing some tests that break on this configuration.

The second commit ("Disable autocorrect when entering text") is an attempt at handling a longstanding issue where controls that have autocorrect enabled on them can lead to unexpected things entered. This is causing a new test failure in the new CI configuration of changing `ll` to `lol`.

I don't see a reason why a test would want to rely on the OS level autocorrect functionality in their test. Please feel free to voice concern about this, looking for opinions. The reasonable (and up-to-now expected) alternative would be to modify the particular text field in the Test App to not have autocorrect enabled.

I still need to do some more testing on this however, as the fix doesn't seem to quite work as expected. I'll mark this "Don't merge yet!" for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: objective-c
 matrix:
   include:
-    - osx_image: xcode6.4
-      env: 'SIMULATOR="name=iPad Air,OS=8.1"'
+    - osx_image: xcode8.3
+      env: 'SIMULATOR="name=iPad Air 2,OS=10.3.1"'
     - osx_image: xcode7.3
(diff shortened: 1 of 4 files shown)


justinseanmartin:
This is also needed for Xcode 9 & iOS 11

justinseanmartin:
#1012 is tracking making a proper fix to this issue, but I'm going to hack a fix in here for now to avoid the autocorrection issue.

mikelupo:
Go ahead and merge it. An attempt to bring us up to Xcode 9 is valid and necessary.

PR #976, opened 2017-05-17
Prefer solution without inspector UI being shown when possible

justinseanmartin:
I'm just reversing the order of these to show the preference to not show the Accessibility Inspector UI in the simulator whenever possible.


justinseanmartin:
Apparently this breaks Xcode 6 builds with iOS8 (getting stuck on `BackgroundTests` for some reason in CI). Need to investigate this further before merging. Maybe I should just give this up until we officially drop Xcode 6 support.

justinseanmartin:
Going to close this for now. I don't see a strong incentive to landing this given that it causes issues with Xcode 6 per the Travis results. Maybe we should revisit that after deprecating Xcode 6 at some point in the future.

PR #981, opened 2017-06-20
Handling the correct error

webventil:
The if clause does not check for `directoryCreationError` which is used for `NSFileManager` to create the directory.


justinseanmartin:
This seems like a no-brainer to merge. Sorry this took so long to get back to! I kicked travis on the one red job to make sure it goes green, and will then merge. Thanks for the contribution!

PR #985, opened 2017-07-05
Added option to select UIPickers based on accessibility identifier.

sebastianv1:


Classes/KIFUIViewTestActor.m
```diff
@@ -345,17 +346,23 @@ - (void)setSwitchOn:(BOOL)switchIsOn;
 
 - (void)selectPickerViewRowWithTitle:(NSString *)title;
 {
-    [self.actor selectPickerViewRowWithTitle:title];
+    KIFUIObject *found = [[self _usingExpectedClass:[UIPickerView class]] _predicateSearchWithRequiresMatch:NO mustBeTappable:NO];
```
justinseanmartin: Should these be `_predicateSearchWithRequiresMatch:YES`?

justinseanmartin:
Looks like the `PickerTests_ViewTestActor` are failing in Travis. Can you take a look at these and see if you can figure out why they are failing?

justinseanmartin:
Thanks for the pull request! Once the feedback is addressed and the tests are passing, we'll merge this.

KIF Tests/PickerTests_ViewTestActor.m
```diff
@@ -107,7 +107,7 @@ - (void)testSelectingRowInComponent
     [[viewTester dateSelector] tap];
     NSArray *date = @[ @"December", @"31", @"2030" ];
     [viewTester selectDatePickerValue:date];
-    [viewTester selectPickerViewRowWithTitle:@"17" inComponent:1];
+    [viewTester selectDatePickerViewRowWithTitle:@"17" inComponent:1];
     [[[viewTester dateSelector] usingValue:@"Dec 17, 2030"] waitForView];
```
sebastianv1: @justinseanmartin Made those changes you suggested. This test was failing because with the new changes inside `KIFUIViewTestActor`, the var `viewTester` would expect a picker view and not a date picker based on the expected class predicate. We decided the least complex and most clear change would be to add this function `selectDatePickerWithRowWithTitle:inComponent` to change individual components in a date picker. Changing individual components of regular pickers would use the original  `selectPickerViewRowWithTitle :inComponent`. Does this look okay to you?

justinseanmartin:
Thanks for the contribution!

Issue #733, opened 2015-10-29
Build error when running c++ tests in KIF


snoopying:
System macro is causing issues when running C++ tests in KIF 

In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/locale:192:

/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/cstdlib:131:9: error: expected unqualified-id

using ::system;

```
    ^
```

In file included from /Users/test/Documents/iOS/Tests/test.mm:19:

In file included from /Users/test/Library/Developer/Xcode/DerivedData/frxusjxqwcj/Build/Products/Debug-iphonesimulator/include/KIF/KIF.h:12:

/Users/test/Library/Developer/Xcode/DerivedData/fwcj/Build/Products/Debug-iphonesimulator/include/KIF/KIFSystemTestActor.h:13:16: note: expanded from macro 'system'
# define system KIFActorWithClass(KIFSystemTestActor)

```
           ^
```

In file included from /Users/test/Documents/iOS/Tests/test.mm:19:

In file included from /Users/test/Library/Developer/Xcode/DerivedData/frxzxqwcj/Build/Products/Debug-iphonesimulator/include/KIF/KIF.h:10:

/Users/test/Library/Developer/Xcode/DerivedData/frxusqwcj/Build/Products/Debug-iphonesimulator/include/KIF/KIFTestActor.h:18:34: note: expanded from macro 'KIFActorWithClass'
# define KIFActorWithClass(clazz) [clazz actorInFile:[NSString stringWithUTF8String:**FILE**] atLine:**LINE** delegate:self]

```
                             ^
```

1 error generated.

Also reported here...

https://groups.google.com/forum/#!topic/kif-framework/qgWxVL644DU 

Is there a workaround for this issue or will this be fixed?

Thanks

mikelupo:
Tony/Brian, I don't know this area of KIF code at all, but I'm curious to know what we can do about the issue.


Thanks,
Mike





-----Original Message-----

phatmann:
Clearly using macros with no prefix at global scope is a terrible idea. At the very least we could rename the macros to  KIF_system and KIF_tester. But really we should get rid of them entirely. We need to anyway for Swift compatibility.

Tony

bnickel:
[...]

PR #988, opened 2017-07-31
Fix cstdlib collisions in KIFSystemTestActor

voznesenskym:
Closes https://github.com/kif-framework/KIF/issues/733


Classes/KIFSystemTestActor.h
```diff
@@ -10,7 +10,11 @@
 #import "KIFTestActor.h"
 #import <UIKit/UIKit.h>
 
+#ifdef __cplusplus
+#define systemTester KIFActorWithClass(KIFSystemTestActor)
+#else
 #define system KIFActorWithClass(KIFSystemTestActor)
```
justinseanmartin: Maybe we should instead do:

```
#define systemTester KIFActorWithClass(KIFSystemTestActor)
#ifndef __cplusplus
// The symbol `system` collides with the cstdlib for compiling C++. Leaving it available for compatibility reasons.
// This will be removed with the next major KIF release, please start using `systemTester` instead.
#define system KIFActorWithClass(KIFSystemTestActor)
#endif
```

That way, we can align around a single definition in the future and switch over all KIF usages from `system` to `systemTester`. Thoughts?

justinseanmartin:
Thanks for the contribution! I've got one comment that I think will make this better for the future. 

If you have time to switch things over to using `systemTester`, that would be great, but I'd be happy to do that after merging as well. 

I've rekicked the Travis CI jobs to get them to go green.

Classes/KIFSystemTestActor.h
voznesenskym: Perfect :) I'll add that

justinseanmartin:
Going to close this, addressing the feedback in #1006.

PR #993, opened 2017-08-11
Add a matcher that allows entering text without validation

justinseanmartin:


justinseanmartin:
@RoyalPineapple @mikelupo - Would you mind taking a look when you get a chance?

PR #994, opened 2017-08-11
Expand the set of exposed UIViewTester matchers and actions

justinseanmartin:
I found some actions that existed in `KIFUITestActor` that weren't being exposed in `KIFUIViewTestActor`. This tries to bring it to parity. The typing into first responder I solved by introducing `usingFirstResponder` as a matcher.

### DIFF
--- a/Classes/KIFUIViewTestActor.m
+++ b/Classes/KIFUIViewTestActor.m
@@ -90,7 +90,6 @@ NSString *const inputFieldTestString = @"Testing";
     return [self usingPredicate:predicate];
 }
 
-
 - (instancetype)usingIdentifier:(NSString *)accessibilityIdentifier;
 {
     NSPredicate *predicate = [NSPredicate predicateWithBlock:^BOOL(id evaluatedObject, NSDictionary *bindings) {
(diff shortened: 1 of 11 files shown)


justinseanmartin:
@RoyalPineapple @mikelupo - This one too when you get a sec.

justinseanmartin:
Fix for the Travis failure requires rebasing the fix on top of #993 after it is merged.

Classes/KIFUIViewTestActor.m
```diff
@@ -263,9 +283,10 @@ - (void)enterTextIntoCurrentFirstResponder:(NSString *)text fallbackView:(UIView
 - (void)setText:(NSString *)text;
 {
     KIFUIObject *found = [self _predicateSearchWithRequiresMatch:YES mustBeTappable:NO];
```
mikelupo: Can *found be nil in this case and if so is it safe to use?
Getting rusty in obj c these days...

justinseanmartin: Yea, ObjC doesn't really have any sort of nullability enforcement, it just relies on nil messaging semantics and assumes that you're ok with everything being nullable.

justinseanmartin: I had messed up the implementation of `setText:` though, so I just pushed a fix for it.

mikelupo:
Thanks Justin. I see no other issues

justinseanmartin:
Exposed one more action for `KIFUIViewTestActor` that we needed. Also, wrapping all the actions in an `autoreleasepool` to prevent the UIView lifetimes from being extended beyond the lifetime of the action.

justinseanmartin:
Rebased with the changes from #993 now merged.

PR #996, opened 2017-08-16
Add an optional compiler flag DEPRECATE_KIF_TESTER

justinseanmartin:
We've deprecated all usage of `tester` and would like to prevent further usages from creeping in.

In order to use it, we are setting: `OTHER_CFLAGS = "$(inherited) -DDEPRECATE_KIF_TESTER=1"`

When that CFLAG is present, it will generate a deprecation warning with the message:
> Use of `tester` has been deprecated; Use `viewTester` instead.


mikelupo:
Looks good to me Justin. Approved.

PR #998, opened 2017-08-16
KIF fails in "selectPickerViewRowWithTitle"

deepakkumarsharma-tudip:
KIF fails in "selectPickerViewRowWithTitle" if UIPickerView having a custom label for title view

### DIFF
--- a/Classes/KIFUITestActor.m
+++ b/Classes/KIFUITestActor.m
@@ -658,9 +658,14 @@
             } else if ([pickerView.delegate respondsToSelector:@selector(pickerView:viewForRow:forComponent:reusingView:)]) {
                 // This delegate inserts views directly, so try to figure out what the title is by looking for a label
                 UIView *rowView = [pickerView.delegate pickerView:pickerView viewForRow:currentIndex forComponent:i reusingView:nil];
-                NSArray *labels = [rowView subviewsWithClassNameOrSuperClassNamePrefix:@"UILabel"];
-                UILabel *label = (labels.count > 0 ? labels[0] : nil);
-                rowTitle = label.text;
+                if ([rowView isKindOfClass:[UILabel class]]) {
+                    UILabel *label = (UILabel *) rowView;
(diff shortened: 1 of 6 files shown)


mikelupo:
I think we may need to add a new test case in order to continue verification of the legacy behavior. @justinseanmartin, @RoyalPineapple  do you agree with this assertion?

justinseanmartin:
Agree that it would be nice to add a test to make sure that we don't regress this in the future. @deepakkumarsharma-tudip - do you think you'd have time to add a regression test to `PickerTests`?

It looks like the same fix is used in `selectPickerValue:pickerType:withSearchOrder:`. We should do some cleanup of those two methods, as it looks like there is a lot of overlap with not 100% consistent behavior (for example supporting `pickerView:attributedTitleForRow:forComponent:` in one but not the other).

deepakkumarsharma-tudip:
@justinseanmartin Yes, I will write it.

justinseanmartin:
Unfortunately, the changes you made to `MainStoryboard.storyboard` fail to be opened on Xcode 6 and Xcode 7, which is why Travis is broken.

It is probably about time that we drop Xcode 6 & 7 support. Users can pull older versions of KIF if they need support for older versions of Xcode.

In order to do so though, we need to get CI working on Xcode 8 and Xcode 9. I'll see if I can find time to do that this week.

justinseanmartin:
Thanks for fixing the storyboard! Will merge on green.

deepakkumarsharma-tudip:
@justinseanmartin I was unable to get the actual cause exactly why build 898.2 is failing, can you please help me out to find the actual cause so that I can fix that ASAP.

justinseanmartin:
I kicked the build off again. I think the machines in Travis sometimes get stuck in a bad state and lead to that exit code 65.

justinseanmartin:
Thanks again!

deepakkumarsharma-tudip:
Thanks a lot guys!

PR #1003, opened 2017-09-06
Make majorSwipeDisplacement smaller

0xpablo:
On iOS 11, when performing a "full swipe" on a TableView cell, the first action is performed by default.

The `majorSwipeDisplacement` is currently used for the amount of displacement perform when swiping a cell and it seems that 0.625 is too much and performs the first action automatically. This is breaking our tests as the action is performed so there is no action to tap anymore.

This PR lowers this constant to 0.5, which I would say it's still a major swipe displacement and does not seem to trigger the new behavior on iOS 11 which would be the expected thing IMO.

Maybe in an additional PR we should also expose a method to perform a full swipe so users are also be able to test this new behavior.

### DIFF
--- a/Additions/UIScreen+KIFAdditions.m
+++ b/Additions/UIScreen+KIFAdditions.m
@@ -11,7 +11,7 @@
 @implementation UIScreen (KIFAdditions)
 
 - (CGFloat)majorSwipeDisplacement {
-    return self.bounds.size.width * 0.625;
+    return self.bounds.size.width * 0.5;
 }
 
 @end
--- a/KIF Tests/Additions/UIScreen+KIFAdditionsTests.m
+++ b/KIF Tests/Additions/UIScreen+KIFAdditionsTests.m
@@ -38,22 +38,22 @@
 
 @implementation UIScreen_KIFAdditionsTests
 
-- (void)test_majorSwipeDisplacement_ScreenSizeEqualTo320Pts_Returns200Pts {
-    CGRect iPhone5PortraitBounds = CGRectMake(0, 0, 320, 468);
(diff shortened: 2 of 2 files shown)


justinseanmartin:
Please fix the assertions in `UIScreen_KIFAdditionsTests` that are failing in CI:
https://travis-ci.org/kif-framework/KIF/jobs/272431409

justinseanmartin:
If you have interest, you could look at updating the Travis config to also run on iOS10 and iOS11 to ensure that it continues working in all configurations.

0xpablo:
Hi @justinseanmartin, It seems I only run the tests on the KIFFrameworkConsumerTests target and not the complete suite. Sorry for that! I fixed the UIScreen_KIFAdditionsTests. 

I was running the whole suite on iOS 11 but there are additional things failing. Such as `testBackgroundApp`, which uses `deactivateAppForDuration`, something a bit hacky that calls into a private API that I would say is broken on iOS 11, and some other tests with `UIActivityViewController`.

I'll try to find some time to check this with the final iOS 11 release to see if something changed, or maybe we'll need to find another workaround or exclude that test on newer versions.

Thanks!

justinseanmartin:
Please rebase to latest master and squash the commits into a single one. 

I just merged a commit that should hopefully make Travis CI work on the Xcode 6 machines again, so that should hopefully bring us green builds to merge.

justinseanmartin:
Also, thanks for fixing the tests!

0xpablo:
Done 🙂, OS X machines on Travis are suffering an outage but once it's fixed it should be good to go. Thanks for your help!

justinseanmartin:
Thanks for the contribution!

PR #1010, opened 2017-09-22
Fix waitForView cannot find UITableViewCell or label, and results in UITableView on iOS 11

banjun:
On Xcode 9 & iOS 11, `waitForViewWithAccessibilityLabel: @"label text on the cell"` returns its `UITableView`. As commented in the existing code, UITableView is too far in view hierarchy to work with cells level.
On iOS 10, it correctly returns `UITableViewCell` or `UITableViewLabel`.

This PR adds testcases to reproduce issue. (After my some inspections, `tapViewWithAccessibilityLabel` change some unknown states which affects reproducibility, I have added 2 testcases.)
Then this patches the view finding logic by searching `-tableViewCell` in addition to `-view`.

`element accessibilityContainer` for table cells is like below:

iOS 10: `UITextAccessibilityElement -> accessibilityContainer = UITableViewCell`
iOS 11: `UITextAccessibilityElement -> accessibilityContainer = UITableViewCellAccessibilityElement -> accessibilityContainer = UITableView`

Thus, we can take `UITableViewCell` on iOS 10, but we unexpectedly go through to `UITableView` on iOS 11.

### DIFF
--- a/KIF Tests/TableViewTests.m
+++ b/KIF Tests/TableViewTests.m
@@ -83,6 +83,20 @@
     [tester tapViewWithAccessibilityLabel:@"Cell 32"];
 }
 
+- (void)testWaitingRowByLabel
+{
+    UIView *v = [tester waitForViewWithAccessibilityLabel:@"First Cell"];
+    XCTAssertTrue([v isKindOfClass:[UITableViewCell class]] || [v isKindOfClass:NSClassFromString(@"UITableViewLabel")], @"actual: %@", [v class]);
+}
+
+- (void)testWaitingRowByLabelAfterTapping
+{
+    // for view lookup changes caused by tapping rows
+    [tester tapViewWithAccessibilityLabel:@"First Cell"];
+    UIView *v = [tester waitForViewWithAccessibilityLabel:@"First Cell"];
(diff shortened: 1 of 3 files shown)


designatednerd:
Guh, thank you for the detailed explanation, it explains some really odd failures I was seeing (entire table view getting swiped off to the side when trying to swipe to open the delete functionality).

KIF Tests/TableViewTests.m
```diff
@@ -83,6 +83,20 @@ - (void)testTappingRowUnderToolbarByLabel
     [tester tapViewWithAccessibilityLabel:@"Cell 32"];
 }
 
+- (void)testWaitingRowByLabel
```
justinseanmartin: Please add the equivalent tests for `TableViewTests_ViewTestActor.m`. We'd like to shift people towards using the new syntax and should have test parity between `viewTester` and `tester`.

justinseanmartin:
Thank you for the contribution and for adding tests that validate the fix! Please squash the commits and address the one piece of feedback and we'll get this merged.

Separately, I'm going to try and get back to having CI working on latest versions of Xcode (#974) so we can ensure this doesn't regress going forward.

banjun:
Added paired testcases to ViewTestActor and squashed.

Issue #1007, opened 2017-09-19
NSInternalInconsistencyException scrolling UITableViews


aeioliu:
Hi there - we're getting this error using the latest version of KIF on any `waitForView(withAccessibilityLabel:)` or `tapView(withAccessibilityLabel:)` that tries to interact with a UITableView:

```
Attempted to dequeue multiple cells for the same index path, which is not allowed. If you really need to dequeue more cells than the table view is requesting, use the -dequeueReusableCellWithIdentifier: method (without an index path). Cell identifier: cell, index path: <NSIndexPath: 0x600000a360a0> {length = 2, path = 0 - 7}
```
The offending code seems to be in `UIView+KIFAdditions.m`, line 254:

`UITableViewCell *cell = [tableView.dataSource tableView:tableView cellForRowAtIndexPath:indexPath];`

It's looking like to us that the issue is related to how that function looping through all the cells in a tableview to find its position, which either fails or takes a considerable amount of time, causing the operation to retry, or the tests to hang/fail. Commenting out this code (lines 252 - 273) allows the tests to progress naturally without interruption. Also, adding a significant amount of time padding (5-10s) for the view to load in with `tester().wait(forTimeInterval:)` before calling `waitForView` or `tapView` seems to help, though this adds significant time to testing.

Even though the code is reused for collectionviews, it doesn't seem like 

Seems like this is only an issue with Xcode 9 and 8.3.3 is unaffected. Stack trace:

```
caught "NSInternalInconsistencyException", "Attempted to dequeue multiple cells for the same index path, which is not allowed. If you really need to dequeue more cells than the table view is requesting, use the -dequeueReusableCellWithIdentifier: method (without an index path). Cell identifier: cell, index path: <NSIndexPath: 0x60c000a3e0e0> {length = 2, path = 0 - 7}"
(
	0   CoreFoundation                      0x000000011374f1cb __exceptionPreprocess + 171
	1   libobjc.A.dylib                     0x00000001184adf41 objc_exception_throw + 48
	2   CoreFoundation                      0x0000000113754362 +[NSException raise:format:arguments:] + 98
	3   Foundation                          0x0000000117f52089 -[NSAssertionHandler handleFailureInMethod:object:file:lineNumber:description:] + 193
	4   UIKit                               0x0000000115b737b4 -[UITableView _dequeueReusableCellWithIdentifier:forIndexPath:usingPresentationValues:] + 454
[...]

PR #1011, opened 2017-09-25
FIX iOS 11 specific bug that causes NSInternalInconsitency Exception

Ralph1300:
Fixes the issue that is also referenced in #1007, causing an exception in a tableview while trying to dequeue cells.

### DIFF
--- a/Additions/UIView-KIFAdditions.m
+++ b/Additions/UIView-KIFAdditions.m
@@ -250,9 +250,19 @@ NS_INLINE BOOL StringsMatchExceptLineBreaks(NSString *expected, NSString *actual
                     }
                     
                     @autoreleasepool {
-                        // Get the cell directly from the dataSource because UITableView will only vend visible cells
-                        UITableViewCell *cell = [tableView.dataSource tableView:tableView cellForRowAtIndexPath:indexPath];
+                        UITableViewCell *cell = [tableView cellForRowAtIndexPath:indexPath];
                         
+                        if (cell == nil) {
+                            @try {
+                                // try to deque the cell, if this fails return false,
+                                // otherwise continue normally
+                                cell = [tableView.dataSource tableView:tableView cellForRowAtIndexPath:indexPath];
+                            }
+                            @catch (NSException * e) {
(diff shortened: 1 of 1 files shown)


justinseanmartin:
I'm a little worried about catching exceptions here for flow control here. It can cause leaks and the fix isn't tightly coupled with the problem we're attempting to address.

If possible, a better fix would be to find a way to avoid the condition that is causing Apple to throw the exception in the first place. I'll try to put some time to this tonight or tomorrow, but would be happy if you wanted to explore in that direction as well.

As an aside, do you know if the same issue exists for UICollectionView?

justinseanmartin:
From my understanding, the existing tests don't hit this issue. Would it be possible for you to add a test case that reproduces the issue that you're addressing? Thanks for contributing!

Ralph1300:
I'm not quite happy with catching the error myself - so if i have time, i will try to figure out why this is happening - another interesting side note (that you already mentioned) is that it is not always happening, i will try to create a sample project to make finding the root cause easier.

Thanks for your input!

mikelupo:
@justinseanmartin @Ralph1300 We use UICollectionViews quite extensively in our code. Not hitting this in the UICollectionView code. Not to say that we won't if this is fixed, but for now...no.

justinseanmartin:
That's good info. We don't hit it in our codebase either, so it seems like it is potentially a bug in how the app is exercising collection views?

knellr:
Just wanted to note that we've been hitting this. I'll be digging a little deeper soon, but it's on transition from a screen with a collection view to a screen with a table view, and we only hit it when the table view has off-screen cells.

justinseanmartin:
Sounds good. If you can get a simple repro case in the KIF integration tests app, I'd be happy to help work on a fix.

mikelupo:
@Ralph1300 have a look at #1020. Maybe you could give it a test run with your own code? As for us, we stopped having the issue and are using it our own automation now.

justinseanmartin:
Thanks for attempting to fix this. I'm going to decline this, as the approach in #1020 is preferable to the approach of catching the exception being thrown.

PR #1024, opened 2017-11-16
Add missing autoreleasepool

jmesmith:
Add an autoreleasepool to prevent object lifetimes from being extended. Thanks @justinseanmartin!


justinseanmartin:
Rekicking Travis to get it green.

PR #1029, opened 2017-12-06
Disable bitcode in podspec

davbeck:
XCTest does not support bitcode. This causes a build failure when running on a device.

### DIFF
--- a/KIF.podspec
+++ b/KIF.podspec
@@ -11,6 +11,7 @@ Pod::Spec.new do |s|
   s.default_subspec         = 'Core'
   s.requires_arc            = true
   s.prefix_header_contents  = '#import <CoreGraphics/CoreGraphics.h>'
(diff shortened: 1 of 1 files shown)


justinseanmartin:
Rekicked the builds due to flakiness. Will merge on green. Thanks for the contribution.

justinseanmartin:
@mikelupo - You test on devices right? Do you not integrate KIF via CocoaPods? I'd have suspected that you'd have hit this before, so just checking...

davbeck:
If you google the error message, you’ll find hundreds of suggestions to turn bitcode off for the entire app. However this solution allows you to have bitcode enabled for the app.

justinseanmartin:
I've rekicked the build 4-5 times and it is still red. I'm going to just merge this for now and will look at seeing if I can make CI more reliable separately.

Issue #889, opened 2016-07-25
Test fails when Accessibility Inspector is off


datinc:
When running our Unit Test on CI environment, test would fail with

`  Assertions: The step timed out after 10.00 seconds: Failed to find accessibility element with the label "Some Label"
  File: SomeClass.m:123`

When I enabled the accessibility inspector using `Settings > General > Accessibility > Accessibility Inspector`, the test all passed.

phatmann:
KIF has code to turn on the accessibility inspector. What version of iOS are you using?

datinc:
Xcode Version 7.3.1
iOS 9.1

phatmann:
Does it work when you run in locally?

datinc:
Locally the same thing is happing
* Passes with accessibility inspector turned on.
* Fails with accessibility inspector turned off.

phatmann:
What KIF version are you using?

datinc:
In my Podfile I have imported it using:
`pod 'KIF/IdentifierTests', '~> 3.0'`

In Podfile.lock you can see that it is 3.4.2
```
  - KIF/Core (3.4.2)
  - KIF/IdentifierTests (3.4.2):
    - KIF/Core
```

phatmann:
There have been a lot of fixes. Can you try the HEAD version? If that works I will push a new release ASAP.

datinc:
On it

datinc:
`pod 'KIF/IdentifierTests', :git => 'https://github.com/kif-framework/KIF.git'`
Still fails.

datinc:
I can reproduce this in a sample project. The steps to reproduce this error:
1. Make sure Accessibility Inspector is turned off on the simulator
2. Re-start the simulator
3. Run test

phatmann:
KIF is supposed to turn on the Accessibility Inspector for you. Can you please debug KIFAcessibilityEnabler to see what is failing?

datinc:
`KIFEnableAccessibility()` is never called in my code. I can see that is because my test case class does not inherit from KIFTestCase. Once I make all my test case classes inherit from this, everything works.

phatmann:
Glad you found the issue.

datinc:
Would this be a bug or a documentation issue. 

> KIFTestCase is not necessary for running KIF tests. Tests can run directly in XCTestCase or any subclass. The basic requirement is that when you call tester or system, self must be an instance of XCTestCase.

It seems to be necessary.

[...]

PR #1033, opened 2017-12-11
Add documentation to call KIFEnableAccessibility in setUp

edwinlai:
Add documentation to call KIFEnableAccessibility in setUp when not inheriting KIFTestCase. Following up on #889 since I ran into the same issue.


justinseanmartin:
Thanks for the documentation update!

PR #1036, opened 2017-12-19
Fix the waitForAnimationToFinishWithTimeout to wait until animation done OR the timeout expires

chibipanda:
Change the wait check to look for the animationKeys count being 0 in the layer. The old check does not do anything (always return true), and as a result, it will just wait until the timeout expires even if the animation finishes much earlier.

### DIFF
--- a/Classes/KIFUITestActor.m
+++ b/Classes/KIFUITestActor.m
@@ -229,7 +229,7 @@ KIFUITestActor *_KIF_tester()
             [window performBlockOnDescendentViews:^(UIView *view, BOOL *stop) {
                 BOOL isViewVisible = [view isVisibleInViewHierarchy];   // do not wait for animations of views that aren't visible
                 BOOL hasUnfinishedSystemAnimation = [NSStringFromClass(view.class) isEqualToString:@"_UIParallaxDimmingView"];  // indicates that the view-hierarchy is in an in-between-state of an animation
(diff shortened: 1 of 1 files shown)


Classes/KIFUITestActor.m
```diff
@@ -229,7 +229,7 @@ - (void)waitForAnimationsToFinishWithTimeout:(NSTimeInterval)timeout stabilizati
             [window performBlockOnDescendentViews:^(UIView *view, BOOL *stop) {
                 BOOL isViewVisible = [view isVisibleInViewHierarchy];   // do not wait for animations of views that aren't visible
                 BOOL hasUnfinishedSystemAnimation = [NSStringFromClass(view.class) isEqualToString:@"_UIParallaxDimmingView"];  // indicates that the view-hierarchy is in an in-between-state of an animation
-                if (isViewVisible && ([view.layer hasAnimations] || hasUnfinishedSystemAnimation)) {
```
justinseanmartin: If you take a look at CALayer-KIFAdditions.m, the `hasAnimations` method is more complicated than just checking the number of `animationKeys`. If there is something wrong with the `hasAnimations` method, it should probably be patched there.

justinseanmartin:
Going to close this PR for now, please feel free to reactivate if you have a response to the above.

PR #1041, opened 2018-01-18
Add -[KIFUIViewTestActor usingAbsenceOfTraits:] 

zadr:
This allows us to differentiate between views that may have the same accessibility label and share some— but not all— accessibility traits.

For example, this allows us to differentiate between a "Next" button in a navigation bar or input accessory (with the ax trait of `.Button`), and a return key that also says "Next" on the input view itself (that may have the ax trait of `.Button` on some OS's \*cough\*iOS 10.2\*cough\* and will always have the ax trait of `.KeyboardKey`).

### DIFF
--- a/Classes/KIFUIViewTestActor.h
+++ b/Classes/KIFUIViewTestActor.h
@@ -75,6 +75,7 @@ extern NSString *const inputFieldTestString;
 /*!
  @abstract Adds a check for accessibility traits to the tester's search predicate.
  @discussion The tester will evaluate accessibility elements looking for matching accessibility traits.
+ Note: You cannot assert the lack of accessibility traits by passing in UIAccessibilityTraitsNone.
  @param accessibilityTraits The accessibility traits of an element to match.
  @return The message reciever, these methods are intended to be chained together.
  */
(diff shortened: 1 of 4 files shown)


Classes/KIFUIViewTestActor.h
```diff
@@ -80,6 +80,14 @@ extern NSString *const inputFieldTestString;
  */
 - (instancetype)usingTraits:(UIAccessibilityTraits)accessibilityTraits;
 
+/*!
+ @abstract Adds a check to avoid views with accessibility traits to the tester's search pedicate.
```
justinseanmartin: NIT: Typo `pedicate`

Classes/KIFUIViewTestActor.m
```diff
@@ -111,6 +111,14 @@ - (instancetype)usingTraits:(UIAccessibilityTraits)accessibilityTraits;
     return [self usingPredicate:predicate];
 }
 
+- (instancetype)usingAbsenceOfTraits:(UIAccessibilityTraits)accessibilityTraits;
+{
+    NSPredicate *predicate = [NSPredicate predicateWithFormat:@"(accessibilityTraits & %@) != %@", @(accessibilityTraits), @(accessibilityTraits)];
```
justinseanmartin: If a bitmask of `usingAbsenceOfTraits:(UIAccessibilityTraitButton | UIAccessibilityTraitSelected)` is supplied, must both traits be absent or if either trait is absent is that sufficient? I'm pretty sure this will match something that doesn't match either of the traits, not both of them. I think I'd expect neither trait to be present though, but I could be convinced maybe. The alternative would be `(accessibilityTraits & %@) == 0`.

Either way, we should make it explicit in the header comment about which behavior is expected. It would also be good to clarify in this methods header and `usingTraits:` header that `usingTraits:UIAccessibilityTraitNone` does nothing. If you don't mind adding that while you're in there, that would be great.

justinseanmartin:
Thanks for adding the test! Overall it looks great, thanks for adding a test for it!

Classes/KIFUIViewTestActor.m
justinseanmartin: We discussed and this is fine as it is. Just added some examples as to the expected behavior in the header.

justinseanmartin:
Going to merge, despite Travis. I believe this will pass given enough time, but Travis is very very very backed up and I don't want to wait for that last job. If it is broken in master, I'll investigate.

Issue #1012, opened 2017-09-29
iOS 11 introduces smart quotes and smart dashes. May cause failure in call to enterText


mikelupo:
Testing on iOS versions prior to 11.0, this is fine,

However when typing into an input field in iOS 11.0,
```
[tester enterText:@"~!@#$%^&*()`1234567890-_+=[{]}:;'\"',<.>/?\\|" intoViewWithAccessibilityIdentifier:@"profile.FirstNameInputField"];
```

the test step fails due to the evaluation that takes place in KIFUITestActor.m
namely, this line fails.
```
KIFTestWaitCondition([actual isEqualToString:expected], error, @"Failed to get text \"%@\" in field; instead, it was \"%@\"", expected, actual);
```

When I look at the actual vs expected in the debugger I see a subtle difference. (namely, the characters just before ",<")
```
**expected:**  "~!@#$^&*()`1234567890-_+=[{]}:;'"',<.>/?\|"

**actual:**    "~!@#$^&*()`1234567890-_+=[{]}:;’”’,<.>/?\|"
```
@justinseanmartin @RoyalPineapple We can make an attempt to figure out what to do with the expected vs actual by doing something like this, 
```
    UITextView *textView = (UITextView *)view;
	if (@available(iOS 11.0, *)) {
		if (textView.smartQuotesType != UITextSmartQuotesTypeNo) //UITextSmartQuotesTypeDefault could pose another risk in that we'd probably have to make some assumptions.
		{
			//Do Something
		}

		if (textView.smartDashesType != UITextSmartDashesTypeNo) //UITextSmartDashesTypeNo could pose another risk in that we'd probably have to make some assumptions.
		{
			//Do Something
		}
	}
```
Discussions? Advice? Nitrous Oxide?

justinseanmartin:
How would you feel about globally disabling `smartDashesType` and `smartQuotesType` on all text fields when running KIF tests? We've done the same internally for `autocorrectionType`, and were considering upstreaming that to avoid those issues that are outside of our control.

Right now, we do it via:
```
    // Prevent autocorrect in all UITextFields

    static dispatch_once_t onceToken;
    dispatch_once(&onceToken, ^{
        struct objc_method_description autocorrectionTypeMethodDescription = protocol_getMethodDescription(@protocol(UITextInputTraits), @selector(autocorrectionType), NO, YES);
        IMP noAutocorrectionTypeIMP = imp_implementationWithBlock(^(UITextField *_self) {
[...]

PR #1044, opened 2018-01-30
Add more generic alert tapping capabilities

startupthekid:
## Added:
- method for tapping system alerts at varying index. This is incredibly useful for tapping system alert (like location alerts) that use new iOS 11 permissions (never allow, allow while in app, always allow - as they pertain to location permissions in this instance).
- refactored `acknowledgeSystemAlert` to call `acknowledgeSystemAlertWithIndex:`

Inspired by: https://github.com/jamesjn/KIF/blob/3361a50389309326aa4a4fc580d4a37421829b58/Classes/UIAutomationHelper.m#L131 (@jamesjn)

### DIFF
--- a/Classes/KIFUITestActor.h
+++ b/Classes/KIFUITestActor.h
@@ -579,6 +579,13 @@ typedef NS_ENUM(NSUInteger, KIFPullToRefreshTiming) {
  @discussion Use this to dissmiss a location services authorization dialog or a photos access dialog by tapping the 'Allow' button. No action is taken if no alert is present.
  */
 - (BOOL)acknowledgeSystemAlert;
+
+/*!
+ @abstract If present, dismisses a system alert with the button at the given index, if any exists, usually 'Allow'. Returns YES if a dialog was dismissed, NO otherwise.
+ @discussion Use this to dissmiss a location services authorization dialog or a photos access dialog by tapping a specified button. No action is taken if no alert is present.
+*/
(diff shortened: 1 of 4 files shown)


justinseanmartin:
Seems reasonable. Thanks for the contribution!

startupthekid:
Why would you merge this? The tests were failing...

startupthekid:
@justinseanmartin why, why, why did you merge when there were failing tests? It wasn't ready to merge because I haven't had time to take a look at the failing CI builds.

justinseanmartin:
Thanks for the concern! I looked at the failing tasks logs and the only thing I saw failing was caused by #1046. I could've asked for a rebase on latest master to get a green build, but the tests had all passed. Master is green:

https://travis-ci.org/kif-framework/KIF/

startupthekid:
I wouldn't have minded a re-base, I'd rather do it correctly than quickly.

startupthekid:
I'm curious too, theses tests are running on XCode 6.4 & 7.3, as well as iOS 8.1, 9.3, and 8.4. How do you know that the tests also pass on later versions of XCode (like 9.2, the most current) and later OS versions?

justinseanmartin:
Simply just an issue of having only two maintainers and very little time to focus on fixing issues that arise. #974 has been languishing waiting to introduce an API to disable autocorrect, which is tracked by #1012. I think we might have someone working on this shortly, but in the meantime I'm going to disable autocorrect on the field that is breaking the test.

PR #1051, opened 2018-02-06
Crash on using a test actor unless KIFEnableAccessibility has already been called

### LINKED ISSUE #1050: viewTester().usingLabel does not work in Simulator iOS 11.2
See attached project.

To reproduce: build and test on iOS Simulator 11.2 in Xcode 9.2. The test will fail.

Do the same on iOS Simulator 11.1 in Xcode 9.2 and the test works.

This is a really simple UIButton, nothing particular to it, just a basic sample project from the Xcode templates, add a UIButton and that's it.
[kiftest.zip](https://github.com/kif-framework/KIF/files/1696828/kiftest.zip)

justinseanmartin:
Resolves #1050

### DIFF
--- a/Classes/KIFAccessibilityEnabler.h
+++ b/Classes/KIFAccessibilityEnabler.h
@@ -9,6 +9,12 @@
 #import <Foundation/Foundation.h>
 
 /**
- * Provides way to enable the Accessibility Inspector.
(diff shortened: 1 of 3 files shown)


justinseanmartin:
@ristkof - Please take a look.

RoyalPineapple:
This is great. Thanks!

PR #1052, opened 2018-02-08
Added ability to enable/disable autocorrect/smart quotes/smart dashes

### LINKED ISSUE #1012: iOS 11 introduces smart quotes and smart dashes. May cause failure in call to enterText

harleyjcooper:
Resolves #1012

This disables autocorrect (as well as smart quotes/smart dashes for iOS 11+) by default, allowing the user to enable/disable them on a per-suite basis (at the minimum level of granularity). I've added some tests to ensure that this behavior is maintained.

This is done by swizzling `UITextField`'s `autocorrectionType`, `smartQuotesType` and `smartDashesType` properties to always return `UITextAutocorrectionTypeNo` `UITextSmartQuotesTypeNo`, and `UITextSmartDashesTypeNo`, respectively. This is entirely reversible, as well, as is demonstrated in the UI tests added.

**KIF Framework Changes:**
- Autocorrect, smart quotes and smart dashes are now disabled by default. In order to re-enable them, override the test suite's `+(void)setUp` method, and call `[KIFTestActor setEnableAutocorrect:YES]`, `[KIFTestActor setEnableSmartQuotes:YES]` or `[KIFTestActor setEnableSmartDashes:YES]`

**Test Changes:**
- Ensure that dashes are modified when smart dashes are enabled, as well as that they're not modified when smart dashes are disabled.
- Ensure that quotes are modified when smart quotes are enabled, as well as that they're not modified when smart quotes are disabled.
- Ensure that autocorrect works when it's enabled, and that it doesn't when it's disabled.


Classes/KIFTextInputTraitsOverrides.h
```diff
@@ -0,0 +1,12 @@
+//
+//  KIFTextInputTraitsOverrides.h
+//  KIF
+//
+//  Created by Harley Cooper on 1/31/18.
+//
+
+#import <Foundation/Foundation.h>
+
+FOUNDATION_EXTERN void KIFSetAutocorrect(BOOL);
```
justinseanmartin: Might be more clear to name these `KIFForceDisableAutocorrect`?

justinseanmartin: Same goes for the other two as well.

justinseanmartin: Nevermind, I see what you're doing (as it is disabled by default)

justinseanmartin: Maybe `KIFAllowDefaultAutocorrectBehavior()` might be more clear, but it is definitely a nit.

justinseanmartin: NIT: Would be slightly more clear as `KIFAllowDefaultAutocorrectBehavior(BOOL)` and same for the two below

KIF Tests/AutocorrectTests.m
```diff
@@ -0,0 +1,57 @@
...
+
+- (void)afterEach
+{
+    [tester tapViewWithAccessibilityLabel:@"Test Suite" traits:UIAccessibilityTraitButton];
+}
+
+- (void)testClearingAndEnteringTypoIntoViewWithAccessibilityLabel
+{
+    [[tester validateEnteredText:NO] clearTextFromAndThenEnterText:@" teh " intoViewWithAccessibilityLabel:@"Greeting"];
+    [[viewTester usingValue:@" teh "] waitForAbsenceOfView];
```
justinseanmartin: Shouldn't use `viewTester` in the non-viewTester tests.

Test Host/TapViewController.m
```diff
@@ -28,9 +28,6 @@ - (void)viewDidLoad
     self.lineBreakLabel.accessibilityLabel = @"A\nB\nC\n\n";
 	self.stepper.isAccessibilityElement = YES;
 	self.stepper.accessibilityLabel = @"theStepper";
-    
-    // Prevent autocorrect from kicking in, breaks tests starting with iOS10
-    self.greetingTextField.autocorrectionType = UITextAutocorrectionTypeNo;
```
justinseanmartin: 👍 thanks for removing the hack

Classes/KIFTestCase.m
```diff
@@ -65,6 +66,11 @@ + (NSArray *)testInvocations
 + (void)setUp
 {
     KIFEnableAccessibility();
+
+    KIFSetAutocorrect([KIFTestActor autocorrectEnabled]);
```
justinseanmartin: Does this need to be called again for each test class?

KIF Tests/AutocorrectTests_ViewTestActor.m
```diff
@@ -0,0 +1,60 @@
...
+
+
+@implementation AutocorrectTests_ViewTestActor
+
++ (void)setUp
+{
+    [KIFTestActor setEnableAutocorrect:YES];
+    [KIFTestActor setEnableSmartQuotes:YES];
+    [KIFTestActor setEnableSmartDashes:YES];
+    [super setUp];
```
justinseanmartin: It is generally encouraged to call `[super setUp]` first. I know this doesn't work with the current paradigm though. Putting a comment elsewhere with a suggestion.

Classes/KIFTestActor.m
```diff
@@ -169,6 +174,36 @@ + (void)setStepDelay:(NSTimeInterval)newStepDelay;
     KIFTestStepDelay = newStepDelay;
 }
 
++ (void)setEnableAutocorrect:(BOOL)enableAutocorrect;
+{
+    KIFAutocorrectEnabled = enableAutocorrect;
```
justinseanmartin: I think there is a bit of a mixed paradigm here between having the static bool configuration and having the `KIFSetAutocorrect` methods, but only updated at `+setUp`. I think this might have been caused by our offline conversation where I'd pointed at putting this setting in here, but I thought you were going in a slightly different direction. Sorry if I caused confusion here.

Thoughts on having these just call directly through to `KIFSetAutocorrect(enableAutocorrect)`? I don't think we actually need the getter or the static BOOL for anything in that case. You could then make the `KIFTestCase` changes be done once in `+[KIFTestActor load]` setting it to `NO` to start, so that it doesn't stomp changes made by tests before they call `+[KIFTestCase setUp]` and it also doesn't require subclassing of `KIFTestCase` if that isn't desired.

I think the tests would stay exactly the same, except you could call `[super setUp]` before or after `setEnableAutocorrect:` and it wouldn't matter.

Classes/KIFTextInputTraitsOverrides.m
```diff
@@ -0,0 +1,75 @@
...
+    if(!autocorrectOriginalImp) {
+        autocorrectOriginalImp = class_getMethodImplementation([UITextField class], @selector(autocorrectionType));
+    }
+
+    IMP autocorrectImp;
+    if(setAutocorrectOn) {
+        autocorrectImp = autocorrectOriginalImp;
+    } else if(!setAutocorrectOn) {
+        autocorrectImp = imp_implementationWithBlock(^(UITextField *_self) {
+            return UITextAutocorrectionTypeNo;
```
justinseanmartin: Swizzling methods isn't a super cheap operation in the runtime. It'd save a little time (not really so bad, milliseconds per test) if you instead replaced each implementation once, and within the `implementationWithBlock` check the value of the static BOOL to return `TypeNo` if it is force disabled, otherwise returns `autocorrectOriginalImp(self, @selector(autocorrectionType))`.

In that case, this could become `KIFSwizzleTextFieldAutocorrect` (that gets called once at test loading time) to do the method replacement, and then expose a `KIFAllowAutocorrectDefaultBehavior(BOOL allowDefaultBehavior)` that just needs to set the static BOOL. The static BOOLs would default to NO, so you don't ever need to call `KIFAllowAutocorrectDefaultBehavior(NO)` in the framework (that's just how it would be by default on being swizzled).

Seems cheaper and easier to understand to me, but up to you.

Classes/KIFTextInputTraitsOverrides.m
```diff
@@ -0,0 +1,75 @@
...
+
+void KIFSetAutocorrect(BOOL setAutocorrectOn) {
+    if(!autocorrectOriginalImp) {
+        autocorrectOriginalImp = class_getMethodImplementation([UITextField class], @selector(autocorrectionType));
+    }
+
+    IMP autocorrectImp;
+    if(setAutocorrectOn) {
+        autocorrectImp = autocorrectOriginalImp;
+    } else if(!setAutocorrectOn) {
```
justinseanmartin: NIT: the `if` on the else here isn't necessary.

KIF Tests/AutocorrectTests.m
harleyjcooper: Whoops. Artifact of some earlier logic.

KIF.xcodeproj/project.pbxproj
```diff
@@ -1442,7 +1462,7 @@
 				DYLIB_CURRENT_VERSION = 1;
 				DYLIB_INSTALL_NAME_BASE = "@rpath";
 				ENABLE_BITCODE = NO;
-				ENABLE_STRICT_OBJC_MSGSEND = YES;
+				ENABLE_STRICT_OBJC_MSGSEND = NO;
```
justinseanmartin: I don't think we'll be able to merge with this disabled. Looking in CocoaPods, this is turned on by default, so it will not pass a `pod lib lint`.

Classes/KIFUITestActor.m
```diff
@@ -153,6 +153,11 @@ - (void)waitForAbsenceOfViewWithAccessibilityLabel:(NSString *)label traits:(UIA
     [self waitForAbsenceOfViewWithAccessibilityLabel:label value:nil traits:traits];
 }
 
+- (void)waitForAbsenceOfViewWithValue:(NSString *)value
+{
+    [self waitForAbsenceOfViewWithAccessibilityLabel:nil value:value traits:UIAccessibilityTraitNone];
+}
+
```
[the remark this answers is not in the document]
justinseanmartin: Agree, this was being used in a test and was then removed. I don't think we need to add this now.

FWIW - I'm glad that this just works possible in the `viewTester` API without needing to add new methods like this.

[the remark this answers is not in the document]
harleyjcooper: Ah, I left a comment and it didn't submit.

It's staleness that's been since removed.

Thanks!

justinseanmartin:
Looks great, thanks for the additional revs on this PR!

PR #1059, opened 2018-03-07
Allow disabling UITestActor animations to speed up tests

pbartolome:
The `KIFUITestActor` by default, performs actions with animation. Those animations (scrolling a table view or a scroll view) can't be speeded up using the `layer.speed` and in a test suite with hundred of tests involving UITableViews this could slow down dramatically the overall running time. 

This PR creates a new settable property in the `KIFUITestActor` to allow disabling this default behaviour.

### DIFF
--- a/KIF Tests/ScrollViewTests.m
+++ b/KIF Tests/ScrollViewTests.m
@@ -16,12 +16,14 @@
 
 - (void)beforeEach
 {
+    XCTAssertTrue([[tester class] testActorAnimationsEnabled]);
     [tester tapViewWithAccessibilityLabel:@"ScrollViews"];
 }
 
 - (void)afterEach
 {
     [tester tapViewWithAccessibilityLabel:@"Test Suite" traits:UIAccessibilityTraitButton];
+    [[tester class] setTestActorAnimationsEnabled:YES];
 }
 
 - (void)testScrollingToTapOffscreenViews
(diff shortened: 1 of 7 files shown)


justinseanmartin:
Please add a tableview test that flips this property on to ensure it doesn't break anything.

justinseanmartin:
There are calls to scroll the view in `UIAccessibilityElement-KIFAdditions.m` and `UIView-KIFAdditions.m` that always pass `animated:YES`. We likely should make those respect this option there as well.

In both of those cases, we spin the runloop an fixed amount of time (.3s and .5s respectively) that can probably be dramatically decreased if we're not waiting for the animation to complete. Maybe we can use .05s if the animations are disabled?

justinseanmartin:
Thanks for the contribution, this looks like a great improvement! Please take a look at the feedback and  let me know your thoughts.

pbartolome:
@justinseanmartin Thanks for your feedback!, I've updated the pull requests with your suggestions and a few test cases replicating the existing tests but with this flag enabled.

KIF Tests/ScrollViewTests.m
```diff
@@ -16,6 +16,7 @@ @implementation ScrollViewTests
 
 - (void)beforeEach
 {
+    [[tester class] setTestActorAnimationsEnabled:YES];
```
justinseanmartin: It would be best to set this back to YES in `afterEach` instead. You could assert that it is YES at the start of the tests if you want to make sure the default value is correct. When the entire test suite is run, if the last test that is executed happens to be one with animations disabled, it might pollute other tests.

justinseanmartin:
One more minor comment and please squash your commits together and I'd be happy to merge this change in. Thanks for the followup!

KIF Tests/ScrollViewTests.m
pbartolome: Where would you assert if the flag is YES? I put it on `beforeEach` in the scrollView/tableview/collectionView tests, but let me know if you were thinking of a different place

justinseanmartin: That is perfect

justinseanmartin:
Thanks for the contribution!

PR #1060, opened 2018-03-20
Adding stabilization code to waitForAnimationsToFinishWithTimeout

harleyjcooper:
We've found that a race condition can sometimes occur whereby a touch event enqueued on the main queue runloop executes before the UI element it's meant to tap has actually appeared onscreen.

When that happens, KIF can wind up sending UI tap events to a view while it's still in the midst of animating onscreen.

We've determined that enqueuing a simple task (set a boolean to `YES`) on the main thread, and spinning a runloop until its execution just before the end of `waitForAnimationsToFinishWithTimeout`, we can avoid this race condition.

Users can configure how long we wait (or if we wait at all, if set to 0) by setting `KIFTestActor.mainThreadDispatchStabilizationTimeout`, or pass a custom timeout to `[KIFUITestActor waitForAnimationsToFinishWithTimeout:(NSTimeInterval)timeout stabilizationTime:(NSTimeInterval)stabilizationTime mainThreadDispatchStabilizationTime:(NSTimeInterval)mainThreadDispatchStabilizationTime]`

### DIFF
--- a/Classes/KIFTestActor.h
+++ b/Classes/KIFTestActor.h
@@ -92,6 +92,7 @@ typedef void (^KIFTestCompletionBlock)(KIFTestStepResult result, NSError *error)
 @property (nonatomic) NSTimeInterval executionBlockTimeout;
 @property (nonatomic) NSTimeInterval animationWaitingTimeout;
 @property (nonatomic) NSTimeInterval animationStabilizationTimeout;
+@property (nonatomic) NSTimeInterval mainThreadDispatchStabilizationTimeout;
 
 - (instancetype)usingTimeout:(NSTimeInterval)executionBlockTimeout;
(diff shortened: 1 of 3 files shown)


justinseanmartin:
Of note, this won't cause tests to fail if the task enqueued on the main thread exceeds the quota (e.g. stuck main queue), it will just insert a log message in the test results. This means the changes might make tests take slightly longer to execute (a few ms per UI tap action), but it will be good for reliability and has an opt-out.

justinseanmartin:
The Travis failure seems to be because the backgrounding test is failing to come back to the foreground. It works fine for me locally though. I'm inclined to try switching that test to run on another device type and see if it fails in the same way.

PR #1065, opened 2018-04-11
Fix #1064 - Take contentInset into account when scrolling to tap on views

### LINKED ISSUE #1064: KIF doesn't take `contentInset` into account when scrolling to tap on views
If you have a table view with more content than can be seen on one screen, under a toolbar, and try to tap on the last row with KIF, KIF fails to make the row visible.

hiltonc:
Changes:

* In `scrollViewToVisible:animated:`, clamp changes to the `contentOffset` so that KIF doesn't scroll outside of the area a user would be able to.
* Remove the `view.isVisibleInWindowFrame` check, as the clamping change now resolves the same problem that the check was intended to fix, and the check was preventing KIF from attempting to scroll in cases where it still needs to (i.e. view is visible in window frame, but obscured by a translucent toolbar).
* Added checks for a visible toolbar to `testTappingRowUnderToolbarByLabel`, `testTappingRowUnderToolbarByLabelWithoutAnimation`, and `testTappingRowUnderToolbarByLabel`. I had previously added these tests, but they stopped being useful because the toolbar was removed from the Test Host storyboard. These checks should make it more apparent why that toolbar exists.
* Update Pod to 3.7.2

Testing:

* Ran UI tests on iPhone 5s simulator with iOS 10.3.1 and 11.2
    * Note: I did find `testPullToRefreshByAccessibilityIdentifierWithDuration` to be a little flakey. Is that a regression?

### DIFF
--- a/Additions/UIAccessibilityElement-KIFAdditions.m
+++ b/Additions/UIAccessibilityElement-KIFAdditions.m
@@ -171,23 +171,23 @@ MAKE_CATEGORIES_LOADABLE(UIAccessibilityElement_KIFAdditions)
         }
         return nil;
     }
-    
-    if (!view.isVisibleInWindowFrame) {
-        // Scroll the view (and superviews) to be visible if necessary
-        UIView *superview = (UIScrollView *)view;
-        while (superview) {
-            // Fix for iOS7 table view cells containing scroll views
-            if ([superview.superview isKindOfClass:[UITableViewCell class]]) {
-                break;
-            }
+
+    // Scroll the view (and superviews) to be visible if necessary
+    UIView *superview = view;
+    while (superview) {
+        // Fix for iOS7 table view cells containing scroll views
+        if ([superview.superview isKindOfClass:[UITableViewCell class]]) {
+            break;
+        }
+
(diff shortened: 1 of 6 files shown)


hiltonc:
This was a collaborative effort with @sassafrass, who figured out the elegant solution to get this to work on iOS 11.

hiltonc:
Apparently when `testPullToRefreshByAccessibilityIdentifierWithDuration` is run immediately after `testPullToRefreshByAccessibilityIdentifier` there is a timing bug. I've updated the two tests to remove the timing bug by adding a reset refresh control button.

justinseanmartin:
3.7.2 is released

hiltonc:
Thank you!

justinseanmartin:
Whoops, I missed that touching the storyboard file broke the Xcode 7 CI job. I thought CI had already finished and given a green build. Any chance that you'd have time to revert the changes to the storyboard for the minimum Xcode version?

```
▸ Compiling MainStoryboard.storyboard
Testing failed:
	The document “(null)” requires Xcode 8.0 or later.
** TEST FAILED **
```

hiltonc:
Sorry about that! #1066 should do the trick.

PR #1068, opened 2018-04-19
Adding autocorrect/smart quotes/smart dashes swizzling for UITextView

harleyjcooper:
The original change to prevent autocorrect/smart quotes/smart dashes only worked on `UITextField`, but not `UITextView`. This simply adds that change for `UITextView` as well.


justinseanmartin:
Waiting on CI before merging.

justinseanmartin:
3.7.3 has been released

PR #1020, opened 2017-11-02
Fix for "Attempted to dequeue multiple cells for the same index path" in table views with Xcode 9

mikelupo:
Fixes #1007

Trying out a fix for issue #1007.
@justinseanmartin @RoyalPineapple please have a look?

Thanks to Larry Gensch (non-github user at this time) who did all of the heavy lifting.

Thanks,
Mike

PR #1070, opened 2018-05-25
Fix bug with table view elements that are matched but offscreen

justinseanmartin:
This bug has existed for a while, but I hadn't gotten around to fixing it until verifying the fix in #1020, which started hitting this in our app.

The issue is that if a table view cell contains a matching element, and that element is fully offscreen but the tableViewCell is partially onscreen, we won't scroll it into view during the matching operation (in `-[UIView(KIFAdditions) accessibilityElementMatchingBlock:notHidden:]`) and tapping the element will fail. 

The fix here is that right before attempting to tap an element, if it isn't currently tappable on the screen and is contained within one or more UIScrollView's, attempt to reposition the scroll views such that the element to tap is visible on the screen.

### DIFF
--- a/Classes/KIFUITestActor.m
+++ b/Classes/KIFUITestActor.m
@@ -321,6 +321,23 @@ static BOOL KIFUITestActorAnimationsEnabled = YES;
         KIFTestWaitCondition(view.isUserInteractionActuallyEnabled, error, @"View is not enabled for interaction: %@", view);
 
         CGPoint tappablePointInElement = [self tappablePointInElement:element andView:view];
+
+        // If the element isn't immediately tappable, try checking if it is contained within scroll views that can be scrolled to make it tappable.
+        if (isnan(tappablePointInElement.x)) {
+            UIView *container = view;
+
+            do {
+                if ([container isKindOfClass:UIScrollView.class]) {
+                    UIScrollView *containerScrollView = (UIScrollView *)container;
+                    CGRect rect = [view convertRect:view.frame toView:containerScrollView];
+                    [containerScrollView scrollRectToVisible:rect animated:NO];
+                }
+
+                container = container.superview;
+            } while (container != nil);
+
+            tappablePointInElement = [self tappablePointInElement:element andView:view];
+        }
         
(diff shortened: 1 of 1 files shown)


Classes/KIFUITestActor.m
```diff
@@ -321,6 +321,23 @@ - (void)tapAccessibilityElement:(UIAccessibilityElement *)element inView:(UIView
...
+
+            do {
+                if ([container isKindOfClass:UIScrollView.class]) {
+                    UIScrollView *containerScrollView = (UIScrollView *)container;
+                    CGRect rect = [view convertRect:view.frame toView:containerScrollView];
+                    [containerScrollView scrollRectToVisible:rect animated:NO];
+                }
+
+                container = container.superview;
+            } while (container != nil);
```
harleyjcooper: Do you never break out of this? It seems like you're going to keep looping. Unless there's an implicit return in the call to `[containerScrollView scrollRectToVisible:animated:]`?

justinseanmartin: `container = container.superview;` will eventually hit the top of the view hierarchy and it's superview will be nil.

harleyjcooper: So no matter what, we always want to run through the element's entire view hierarchy, calling `[containerScrollView scrollRectToVisible:animated:]` on everything we can? We're okay with potentially calling that multiple times?

justinseanmartin: Yep, we'll call it on all of the scroll views up to the root of the hierarchy. If you have a scrollview within a scrollview, all of them should be positioned such that the element to be tapped on should scroll to be visible.

mikelupo:
I will try to finish this up to apply Justin’s comments today or tomorrow. Sorry this is taking so long.

justinseanmartin:
I'm not sure that this is the right place to fix this. I know there is a bug here, the question is more on what the right way to address it. Going to leave this open for a while longer and hope I get time to actually investigate this.

justinseanmartin:
I think this might actually be fixed now, between the changes in #1102 and #1103. I'm going to close this PR. Please report if you're still seeing this issue once those changes are landed.

PR #1073, opened 2018-06-04
Update XCTestCase-KIFAdditions.m

maryamaffinityclick:
add KIFScreenshot as attachment to XCTContext activity when KIF_SCREENSHOT is set as environment variable. So the report in Xcode or any html report would include the snapshot like below : 
screen shot 2018-06-04 at 4 57 51 pm

### DIFF
--- a/Additions/XCTestCase-KIFAdditions.m
+++ b/Additions/XCTestCase-KIFAdditions.m
@@ -65,12 +65,17 @@ static inline void Swizzle(Class c, SEL orig, SEL new)
 - (void)writeScreenshotForException:(NSException *)exception;
 {
     [[UIApplication sharedApplication] writeScreenshotForLine:[exception.userInfo[@"LineNumberKey"] unsignedIntegerValue] inFile:exception.userInfo[@"FilenameKey"] description:nil error:NULL];
-        
+    
+    dispatch_semaphore_t semaphore = dispatch_semaphore_create(0);
+
     [XCTContext runActivityNamed:(@"screenshot") block:^(id<XCTActivity>  _Nonnull activity) {
         XCUIScreenshot *screenShot = [[XCUIScreen mainScreen] screenshot];
         XCTAttachment *attachment = [XCTAttachment attachmentWithScreenshot:screenShot];
         [activity addAttachment:(attachment)];
+        dispatch_semaphore_signal(semaphore);
     }];
+    
+    dispatch_semaphore_wait(semaphore, DISPATCH_TIME_FOREVER);
 }
 
 - (void)printViewHierarchyIfOptedIn;
--- a/Additions/XCTestCase-KIFAdditions.m
+++ b/Additions/XCTestCase-KIFAdditions.m
@@ -68,12 +68,18 @@ static inline void Swizzle(Class c, SEL orig, SEL new)
     
     dispatch_semaphore_t semaphore = dispatch_semaphore_create(0);
 
+#ifdef __IPHONE_11_0
+    dispatch_semaphore_t semaphore = dispatch_semaphore_create(0);
+
     [XCTContext runActivityNamed:(@"screenshot") block:^(id<XCTActivity>  _Nonnull activity) {
         XCUIScreenshot *screenShot = [[XCUIScreen mainScreen] screenshot];
         XCTAttachment *attachment = [XCTAttachment attachmentWithScreenshot:screenShot];
         [activity addAttachment:(attachment)];
         dispatch_semaphore_signal(semaphore);
     }];
+    
+    dispatch_semaphore_wait(semaphore, DISPATCH_TIME_FOREVER);
+#endif
     
     dispatch_semaphore_wait(semaphore, DISPATCH_TIME_FOREVER);
 }
--- a/Additions/XCTestCase-KIFAdditions.m
+++ b/Additions/XCTestCase-KIFAdditions.m
@@ -67,6 +67,7 @@ static inline void Swizzle(Class c, SEL orig, SEL new)
     [[UIApplication sharedApplication] writeScreenshotForLine:[exception.userInfo[@"LineNumberKey"] unsignedIntegerValue] inFile:exception.userInfo[@"FilenameKey"] description:nil error:NULL];
     
 #ifdef __IPHONE_11_0
+    //semaphore will make sure the screenshot will be captured. otherwise it will crash on getting screenshot!  
     dispatch_semaphore_t semaphore = dispatch_semaphore_create(0);
 
     [XCTContext runActivityNamed:(@"screenshot") block:^(id<XCTActivity>  _Nonnull activity) {
--- a/Additions/XCTestCase-KIFAdditions.m
+++ b/Additions/XCTestCase-KIFAdditions.m
@@ -65,6 +65,12 @@ static inline void Swizzle(Class c, SEL orig, SEL new)
 - (void)writeScreenshotForException:(NSException *)exception;
 {
     [[UIApplication sharedApplication] writeScreenshotForLine:[exception.userInfo[@"LineNumberKey"] unsignedIntegerValue] inFile:exception.userInfo[@"FilenameKey"] description:nil error:NULL];
+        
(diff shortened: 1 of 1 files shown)


maryamaffinityclick:
some checks failed, What is the problem with my changes?

justinseanmartin:
This fails to compile on Xcode 8 & 7. You should be able to wrap this in a `#ifdef __IPHONE_11_0` to avoid the compilation errors (the functionality will only be available on Xcode 9+).

Additions/XCTestCase-KIFAdditions.m
```diff
@@ -65,6 +65,17 @@ - (void)KIF_recordUnexpectedFailureWithDescription:(id)arg1 exception:(NSExcepti
 - (void)writeScreenshotForException:(NSException *)exception;
 {
     [[UIApplication sharedApplication] writeScreenshotForLine:[exception.userInfo[@"LineNumberKey"] unsignedIntegerValue] inFile:exception.userInfo[@"FilenameKey"] description:nil error:NULL];
+    
+    dispatch_semaphore_t semaphore = dispatch_semaphore_create(0);
+
+    [XCTContext runActivityNamed:(@"screenshot") block:^(id<XCTActivity>  _Nonnull activity) {
+        XCUIScreenshot *screenShot = [[XCUIScreen mainScreen] screenshot];
```
justinseanmartin: This requires having the `UIAutomation` framework linked into KIF. We do have this today, but I'm hesitant to add more dependencies on `XCUI`, which we can't use on devices.

At the very least, this needs to be wrapped in a `#if TARGET_IPHONE_SIMULATOR` like the method `acknowledgeSystemAlert` is (for the same reasons).

Additions/XCTestCase-KIFAdditions.m
```diff
@@ -65,6 +65,17 @@ - (void)KIF_recordUnexpectedFailureWithDescription:(id)arg1 exception:(NSExcepti
 - (void)writeScreenshotForException:(NSException *)exception;
 {
     [[UIApplication sharedApplication] writeScreenshotForLine:[exception.userInfo[@"LineNumberKey"] unsignedIntegerValue] inFile:exception.userInfo[@"FilenameKey"] description:nil error:NULL];
+    
+    dispatch_semaphore_t semaphore = dispatch_semaphore_create(0);
```
justinseanmartin: Is `XCTContext runActivityNamed:block:` not synchronous? Nothing in the documentation seems to indicate that it will invoke the block asynchronously:
https://developer.apple.com/documentation/xctest/xctcontext/2887132-runactivitynamed

maryamaffinityclick: I was getting crash sometimes that It couldn't get the screenshot and . I fixed it in this way

Additions/XCTestCase-KIFAdditions.m
maryamaffinityclick: thats right, I will be adding those #if s

maryamaffinityclick: apparantly  the above code for write screenshot doest work on device too and it will crash on 
```             *error = [NSError KIFErrorWithFormat:@"Couldn't create directory at path %@ (details: %@)", outputPath, directoryCreationError];
```  so it doesn't reach my code. 
so if that doesnt work on device I should be checking for target for that one too , am I right ?

Comment (maryamaffinityclick) [reply Additions/XCTestCase-KIFAdditions.m]: actually I didn't quit get what you mean by `having the UIAutomation framework linked into KIF`.  I tried on device and it worked fine in my project.

Comment (justinseanmartin) [reply Additions/XCTestCase-KIFAdditions.m]: Take a look at `-[UIAutomationHelper linkAutomationFramework]` for what I mean. The screenshotting method (and all things relying on XCUI) should probably be exposed as methods from that class.

File: Additions/XCTestCase-KIFAdditions.m (hunk 2 above)
Comment (justinseanmartin) [reply Additions/XCTestCase-KIFAdditions.m]: Interesting, I wonder why that would be. If it isn't a synchronous block invocation, then this makes sense to keep it as is. Please add a comment with some information about what the behavior was you were seeing without the semaphore for the next person that encounters this code.

**Comment by justinseanmartin** [comment]
Travis failure is now on Xcode 9:

```
❌  /Users/travis/build/kif-framework/KIF/Additions/XCTestCase-KIFAdditions.m:72:26: redefinition of 'semaphore'
    dispatch_semaphore_t semaphore = dispatch_semaphore_create(0);
```

Additions/XCTestCase-KIFAdditions.m
maryamaffinityclick: but I do want to have it for devices as well. cause it can be used on devices.  what you are saying is that instead of checking for simulator I can put the screenshot codes in the `UIAutomationHelper` class as a function and call that to get the screenshot  in `XCTestCae-KIFAdditions` ?

justinseanmartin: Yep, that's what I meant for where to put this code.

Regarding having this functionality work on physical iOS devices, unfortunately it can't (except jailbroken devices) with this relying on XCUI for the functionality. This is the portion that will fail:
https://github.com/kif-framework/KIF/blob/2453685d15221a50ba73bc345031d940e864e7be/Classes/UIAutomationHelper.m#L193

It might be worthwhile to file a radar to Apple requesting them to expose this functionality for usage within XCTest.

justinseanmartin: This should be moved to `UIAutomationHelper`, and all public functionality of it should probably be made to either:

1. be compiled out on physical devices (fail at compile time, don't support jailbroken devices)
2. crash on calling one of these methods on a physical device (fail at runtime, support jailbroken devices)
3. store a bool if the call to `linkAutomationFramework` fails and log some message (degrade gracefully, less obvious failure mode)

maryamaffinityclick: Thanks for explanation. I agree that I should move those UI related things to that class and I know that the `acknowledgeSystemAlert` is not supported for actual device except the jailbroken ones. but here in my case it does actually work on non-jailbroken device. even the UIAutomation is disabled it will get the screenshot of the device. so I don't get your point about handling jailbroken and non-jailbroken device ??

justinseanmartin: You're able to run this on a real physical iOS device, running a KIF test as a unit test (not as UI test)? That is surprising to me. I'll have to give this a shot, but I don't generally run our automation on physical devices. I'll try to verify this soon.

maryamaffinityclick: Yes exactly ! KIF as unit test. as I understand these code are all in XCTest framework so we don't have those issue that you are talking about.  it will run fine on simulator and real physical device ( not jailbroken ) !

mikelupo:
Yup,
I’m using 9.2 & up in my lab.
I’ll give this a shot during the work week.

mikelupo:
@maryamaffinityclick I pulled this PR and am running it locally. I do see that a screenshot exists in my app's ~/Documents folder, but I don't see any image of the screenshot anywhere in XCode. 
Can you please advise?

Specifics on what I did:
I set the KIF_SCREENSHOTS=~/Documents in both the KIF scheme and the Test Host scheme. 
I turned off most of the tests (just for brevity).
I introduced a failure by waiting for something that isn't there.
Test fails as I expect.
Screenshot exists in the devices itself for "Test Host" app.
Can't find the same anywhere else in Xcode.  (This is where I'm confused.)

Thanks!
Mike

maryamaffinityclick:
@mikelupo my screenshot is from the "report navigator" in Xcode. When you run the tests in Xcode it will show the report under each target with 3 sections : "Build", "Debug", "Test" . if you click on "Test" you have to see name of your test cases and each test case if they fail they suppose to have an attachment which will be the screenshot .
FYI This is a new feature added in XCTest framework and only works on Xcode 9+ and running on iOS 11+ .

mikelupo:
Thanks @maryamaffinityclick 
I tested this using Xcode 9.4 on an iPhone X (running iOS 11.4) and again on an iPhone 6 (running iOS 10.3.1). 
Seems good to integrate.

Thank you for your contribution to KIF!
Mike

PR #1081, opened 2018-08-02
Fix crash on rotate on very small angle

justinseanmartin:
The issue was initially reported and fixed by @Screon in #1080, I'm building upon their commit.

* [Screon] Don't access out of bounds index in `dragPointsAlongPaths:` with an empty array or points.
* Add a test case that would hit the failure from #1080
* Delete the redundant tests for `KIFUITestActor` (the tests aren't really exercising test actor code).
* Add some generics to the arrays to make it a bit more clear
* Always send at least 2 points for the path in 'twoFingerRotateAtPoint:angle:'.
* Add a bit more safety in 'dragPointsAlongPaths:', bail early if the array of points is empty.


Additions/UIView-KIFAdditions.m
```diff
@@ -571,14 +571,14 @@ - (void)dragAlongPathWithPoints:(CGPoint *)points count:(NSInteger)count;
     [self dragPointsAlongPaths:@[[array copy]]];
 }
 
-- (void)dragPointsAlongPaths:(NSArray *)arrayOfPaths {
-    // must have at least one path, and each path must have the same number of points
-    if (arrayOfPaths.count == 0)
+- (void)dragPointsAlongPaths:(NSArray<NSArray<NSValue *> *> *)arrayOfPaths {
```
harleyjcooper: Should we also be making similar changes to `- (NSArray *)pointsFromStartPoint:(CGPoint)startPoint toPoint:(CGPoint)toPoint steps:(NSUInteger)stepCount`? It's internal to `UIView-KIFAdditions`, but it also returns an `NSArray *` - should that be returning an `NSArray<NSArray<NSValue *> *> *`?

justinseanmartin: Done

PR #1083, opened 2018-08-15
tableview search method optimization

LWX124:


mikelupo:
Hi @LWX124, Thank you for your PR.
Are you aware of this? #1020 I fear that your optimization might bring the "Attempt to dequeue multiple cells for the same index path crash" back.

justinseanmartin:
Can you explain a little bit about what problem you're seeing and what you're trying to accomplish with the change? That will help a lot in evaluating the PR. Thanks!

LWX124:
> 
[Uploading QQ20180816-143310-HD.mp4.zip…]()
Some tests like this. Cell can't be loaded and always splashing while trying to find view

LWX124:
I've send u an email with our recorded video.

justinseanmartin:
I saw the video, thanks for sending it. It looks to be an issue potentially specifically with large table views and possibly the video elements they contain or maybe due to the variable cell height?

@LWX124 - The fix that you provided here fully resolves the issue for you? If so, why decline the PR? Do you not want to go forward with the change?

Both changes seemed reasonable:
* Make the key window always be the first one searched (via reverseObjectEnumerator in `UIApplication-KIFAdditions.m`)
* Only call `scrollRectToVisible:` once per section. 
  * I'm not even sure why that is even necessary at all on a per-section basis, because it will immediately after `scrollToRowAtIndexPath:`. 
  * @mikelupo - do you know if/why the `scrollRectToVisible:` call for the section is necessary?

LWX124:
The fix seems to work well for my case, but it may cause other new issues. KIF works very well for our project before updating, I'm afraid that maybe some commit have conflict with our project. I'm trying to find the real reason and make another PR.

justinseanmartin:
I'm guessing that all the scrolling that is being done in the changes to fix #1007 is expensive and is causes KIF to become very sluggish with your long list of table view cells containing live videos.

Issue #1085, opened 2018-08-20
Please update "Use With Swift" section


datinc:
I was intermittently struggling with getting UIAlertController buttons to be pressed. Tests would randomly fail when I tried to tap the button. I know in the readme you specifically call out that you need to call `KIFEnableAccessibility()` if you are using XCTestCase directly... however in the readme section [Use With Swift](https://github.com/kif-framework/KIF/blob/master/README.md#use-with-swift) you provide code which I happily copied and pasted. You would have saved me more time than I would like to admit if a code snippet was added to ensure that `KIFEnableAccessibility()` was called. I ended up using [XCTestObservationCenter](https://developer.apple.com/documentation/xctest/xctestobservationcenter) and setting NSPrincipalClass to `TestBundleName.ObserverClassName`. Worked like a charm.

justinseanmartin:
Please take a look at the proposed solution and let me know what you think.

PR #1086, opened 2018-08-22
Issue #1085: Make XCTestCase 'just work'

justinseanmartin:
There isn't a need to require test writers to explicitly call
KIFEnableAccessibility. Instead, call to enable accessibility on
the first attempt to perform an action via KIFTestActor.

This resolves issue #1085.


justinseanmartin:
Comment was left over on the commit:
https://github.com/justinseanmartin/KIF/commit/28f5c28d920ee35ab1f1f95d8545c34c5bfb3c17#commitcomment-30256992

PR #1094, opened 2018-11-28
Fix scrollViewToVisible:animated:

plu:
When `UIScrollView` has some `contentInset` set, the `scrollViewToVisible:animated:` is not scrolling to the right position. This PR exposes the bug in the tests and fixes the implementation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,6 @@ matrix:
 before_install:
   - xcrun simctl list
   - brew update || brew update
-  - rvm install 2.3.1
   - gem install bundler
 install:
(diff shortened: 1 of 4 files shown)


justinseanmartin:
Thanks for fixing the CI builds while you were in there, appreciate it!

justinseanmartin:
Want a dot release with the fix?

plu:
Thanks for merging it!

> Want a dot release with the fix?

Not yet, thanks for asking! Still looking into the issue we're discussing in the comments of 786fa8b23c7a1bcbf50999720e92992d780a31a8.

PR #1098, opened 2019-03-17
Fix warning of XCUIScreen availability

kylejm:
In a project of ours that uses KIF, we were receiving a warning for usage of XCUIScreen on [this line](https://github.com/kif-framework/KIF/blob/master/Additions/XCTestCase-KIFAdditions.m#L78).

I've resolved this by adding an `@available` check.

### DIFF
--- a/Additions/XCTestCase-KIFAdditions.m
+++ b/Additions/XCTestCase-KIFAdditions.m
@@ -70,7 +70,7 @@ static inline void Swizzle(Class c, SEL orig, SEL new)
     // For now, this will be disabled and can be fixed later by anyone
     // that relies on this functionality.
 #if defined(__IPHONE_11_0) && !defined(__IPHONE_12_0)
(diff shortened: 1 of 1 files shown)


justinseanmartin:
Looks great, and works at least back to Xcode 8 (verified by Travis) so looks good to me. Thanks for the contribution!

kylejm:
@justinseanmartin awesome! Thank you 😊

What's the release cycle with KIF? We're currently depending on my fork and would love to move back to a release of KIF.

justinseanmartin:
I'd be happy to do a bugfix release for this change. You could also point at the github repo with the specific SHA using `:commit => '8181a77f123'`, but that won't be a problem after I've published a new version. Will update shortly.

justinseanmartin:
Sorry, I got distracted by some other things. 
https://cocoapods.org/pods/KIF -> 3.7.5 has now been published

PR #1100, opened 2019-04-03
Fix issue with elements not becoming first responder fast enough

justinseanmartin:
When an UIPopover is presented in an animated fashion when an element is tapped, that element may not become first responder until after the animation is completed. This change gives waits for the animation to complete before checking that it became first responder. 

Moving the `waitForAnimationsToFinish` introduced some other test failures when the element that became first responder is moved offscreen. For this reason, the additional check of `![view isProbablyTappable]` was added.


justinseanmartin:
It looks like this breaks `WebViewTests testTappingLinks`, so I need to look into that before merging this change.

justinseanmartin:
@harleyjcooper - Please take another look. I had to add an exception for cases like web views. I think this should go green now though.

PR #1102, opened 2019-04-15
Remove check preventing scroll UITableViewCells

justinseanmartin:
It prevents partially offscreen UITableViewCell elements from being made visible. I don't think this clause is necessary with an iOS 8+ minimum supported version we have now for KIF.


harleyjcooper:
LGTM

PR #1103, opened 2019-04-16
Fixing broken table scrolling

harleyjcooper:
When attempting to tap offscreen table elements, after `[tableView scrollToRowAtIndexPath:indexPath atScrollPosition:UITableViewScrollPositionNone animated:animationEnabled]` is called, the scrolling never actually occurs before the call to `[tableView cellForRowAtIndexPath:indexPath]`, which will always return `nil` for offscreen table cells. A yield could be inserted in there, but for large tables, that would be an enormous amount of time.

The spinning of the runloop after that block only allows the final call to `scrollToRowAtIndexPath` to be executed, so the code will only succeed if the cell is in view initially, or it's in the final scrolled section in the table. This code, as it was, would never succeed if the cell being sought was above the initially scrolled area of the table.

Switching the `scrollToRowAtIndexPath` call to `[tableView scrollRectToVisible:sectionRect animated:NO]` is more instantaneous, requires no yield, and works for cells above AND below the currently-scrolled portion of the table.

Plus, the final call to `[tableView scrollRectToVisible:initialPosition animated:NO]` will succeed in moving back to the initial position for all cases, essentially instantaneously.

### DIFF
--- a/Additions/UIView-KIFAdditions.m
+++ b/Additions/UIView-KIFAdditions.m
@@ -242,7 +242,7 @@ NS_INLINE BOOL StringsMatchExceptLineBreaks(NSString *expected, NSString *actual
             }
 
             UITableView *tableView = (UITableView *)self;
-            __block NSIndexPath *firstIndexPath = nil;
+            CGRect initialPosition = CGRectMake(tableView.contentOffset.x, tableView.contentOffset.y, tableView.frame.size.width, tableView.frame.size.height);
 
             // Because of a bug in [UITableView indexPathsForVisibleRows] http://openradar.appspot.com/radar?id=5191284490764288
             // We use [UITableView visibleCells] to determine the index path of the visible cells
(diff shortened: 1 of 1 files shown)


justinseanmartin:
Overall LGTM as long as you update the PR description to explain what you're doing and why.

dflems:
@justinseanmartin Can you cut a new version once this is merged?

justinseanmartin:
Yea, @dflems - were you able to test the change, did it work for you all?

harleyjcooper:
@dflems We've released KIF 3.7.6 - https://github.com/kif-framework/KIF/releases/tag/v3.7.6

dflems:
@justinseanmartin Looking good. Previously ~40 tests were failing/flaking and now it's down to two (looks like its for other reasons than the scrolling though). I'll follow up once we diagnose those.

PR #1106, opened 2019-04-19
Fixing UICollectionView cell tapping

harleyjcooper:
When tapping a `UICollectionView`, sometimes the data in the cell might change prior to being displayed, since we're only pulling it from the data source. So here we simulate the cell appearing onscreen.


harleyjcooper:
> LGTM. Would be nice if there were a test added to the functional test suite to ensure this doesn't regress, but it also isn't the end of the world.

This would take a bit of effort in rearchitecting the `CollectionViewController` - it would need us to implement `collectionView:willDisplayCell:forItemAtIndexPath` instead of `collectionView:cellForItemAtIndexPath`. I created #1107 to track this.

PR #1061, opened 2018-03-26
Add "atPosition:" variations to UICollectionView waitForCellAtIndexPath methods

gaperlinski:
Allows the consumers to specify custom `UICollectionViewScrollPosition` of the cell they want to scroll to.  `UICollectionViewScrollPositionCenteredHorizontally | UICollectionViewScrollPositionCenteredVertically` remains to the default value.

PR #1109, opened 2019-05-22
Add "atPosition:" variations to UICollectionView waitForCellAtIndexPa…

harleyjcooper:
Taking over from @gaperlinski for #1061 :

> Allows the consumers to specify custom UICollectionViewScrollPosition of the cell they want to scroll to. UICollectionViewScrollPositionCenteredHorizontally | UICollectionViewScrollPositionCenteredVertically remains to the default value.

### DIFF
--- a/KIF Tests/CollectionViewTests.m
+++ b/KIF Tests/CollectionViewTests.m
@@ -59,6 +59,27 @@
 }
 
 
+- (void)testTappingLastAndFirstRowAtPositionWithAccessiblityIdentifier
+{
+    UICollectionViewCell *lastCell = [tester waitForCellAtIndexPath:[NSIndexPath indexPathForRow:-1 inSection:-1] inCollectionViewWithAccessibilityIdentifier:@"CollectionView Tests CollectionView" atPosition:UICollectionViewScrollPositionBottom];
+    CGPoint lastPosition = [lastCell.superview convertPoint:lastCell.frame.origin toView:nil];
+    XCTAssertEqual(lastCell.frame.origin.x, lastPosition.x);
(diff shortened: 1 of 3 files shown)


gaperlinski:
Thanks, @harleyjcooper ! I've been pretty lousy with finding find to add the test. Btw, I'm wondering why tests fail on iPad simulator. Doesn't seem to be connected to this change.

KIF Tests/CollectionViewTests.m
```diff
@@ -59,6 +59,27 @@ - (void)testTappingLastItemAndSectionWithoutAnimation
 }
 
 
+- (void)testTappingLastAndFirstRowAtPositionWithAccessiblityIdentifier
```
gaperlinski: I would rename this and the other test to `testScrolling` since you're not really tapping anything

KIF Tests/CollectionViewTests.m
```diff
@@ -59,6 +59,27 @@ - (void)testTappingLastItemAndSectionWithoutAnimation
...
+{
+    UICollectionViewCell *lastCell = [tester waitForCellAtIndexPath:[NSIndexPath indexPathForRow:-1 inSection:-1] inCollectionViewWithAccessibilityIdentifier:@"CollectionView Tests CollectionView" atPosition:UICollectionViewScrollPositionBottom];
+    CGPoint lastPosition = [lastCell.superview convertPoint:lastCell.frame.origin toView:nil];
+    XCTAssertEqual(lastCell.frame.origin.x, lastPosition.x);
```
gaperlinski: Is this a vertically scrolling collection view? If so, shouldn't you be checking here the `y` position (whether it aligns to the top/bottom of the collection view bounds)?

harleyjcooper: OK I'm comparing the y now. I'm a bit fuzzy on UIView coordinate conversion, but this seems to work.

gaperlinski: LGTM

RoyalPineapple:
This all looks totally reasonable to me. Thanks Harley!

PR #1111, opened 2019-06-05
Fix Xcode 11 (Beta) build error (implicit conversion warning)

jneidlinger:
Fixes:

```
KIF/Classes/KIFSystemTestActor.m:61:64: error: implicit conversion loses integer precision: 'UIDeviceOrientation' (aka 'enum UIDeviceOrientation') to 'int' [-Werror,-Wshorten-64-to-32]
    [[UIDevice currentDevice] setValue:[NSNumber numberWithInt:orientation] forKey:@"orientation"];
```

### DIFF
--- a/Classes/KIFSystemTestActor.m
+++ b/Classes/KIFSystemTestActor.m
@@ -58,7 +58,7 @@
 
 - (void)simulateDeviceRotationToOrientation:(UIDeviceOrientation)orientation
 {
-    [[UIDevice currentDevice] setValue:[NSNumber numberWithInt:orientation] forKey:@"orientation"];
+    [[UIDevice currentDevice] setValue:[NSNumber numberWithInt:(int)orientation] forKey:@"orientation"];
 }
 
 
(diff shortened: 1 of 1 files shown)


Classes/KIFSystemTestActor.m
```diff
@@ -58,7 +58,7 @@ - (void)simulateMemoryWarning
 
 - (void)simulateDeviceRotationToOrientation:(UIDeviceOrientation)orientation
 {
-    [[UIDevice currentDevice] setValue:[NSNumber numberWithInt:orientation] forKey:@"orientation"];
+    [[UIDevice currentDevice] setValue:[NSNumber numberWithInt:(int)orientation] forKey:@"orientation"];
```
justinseanmartin: Probably could just use `@(orientation)` and avoid the explicit cast. That'd be cleaner if it works IMO.

justinseanmartin: Tested it locally and it fixes the compiler error. Went ahead and made that change in this branch.

justinseanmartin:
Thanks for calling this out and the contribution @jneidlinger! Will merge this on green builds.

Classes/KIFSystemTestActor.m
jneidlinger: Nice, definitely cleaner. Thanks for the fast follow up!

Issue #1119, opened 2019-08-29
iOS Deprecated API Usage Warning


miquelalvarado:
Just submitted my app to TestFlight and got an email with the error:

ITMS-90809: Deprecated API Usage - Apple will stop accepting submissions of apps that use UIWebView APIs . See https://developer.apple.com/documentation/uikit/uiwebview for more information.

I was wondering if all the code related to UIWebView will have to be removed and only keep WKWebView.

It was just a warning, the build ended up being available on TestFlight, but I think this is something that will have to be addressed at some point.

damirstuhec:
What is the status on this?

ggdiez:
PR is launched, I'm waiting for approved or comments

justinseanmartin:
Apologies for the delay, just put up some PR feedback.

Why is it that your app has KIF symbols linked into it for AppStore submission? I would expect KIF to only be getting linked into test targets.

ximenaperez:
hi! what's the status on this?? it will be deprecated on Dec 2020

PR #1121, opened 2019-09-17
Use WKWebview instead UIWebView because last one is deprecated in iOS 12

ggdiez:
This PR fix issue #1119 

I just changed UIWebView by WKWebview because first one is deprecated in iOS 12 and each time to upload binary to Apple you receive a warning.


damirstuhec:
Bump.

Additions/UIView-KIFAdditions.m
```diff
@@ -458,7 +459,7 @@ - (void)tapAtPoint:(CGPoint)point;
     
     if ([NSStringFromClass([self class]) isEqual:@"UIWebBrowserView"]) {
         webBrowserView = self;
-    } else if ([self isKindOfClass:[UIWebView class]]) {
+    } else if ([self isKindOfClass:[WKWebView class]]) {
```
justinseanmartin: Does it make sense to maintain some `UIWebView` support side-by-side until Apple starts rejecting app store submissions (I assume will likely be in the Spring)?

ggdiez: Hi @justinseanmartin 
WKWebview is available since iOS 8 like KIF project. Sooner people has to move to WKWebview and they can because is supported.

We can keep both something like:
`else if ([self isKindOfClass:[UIWebView class]] || [self isKindOfClass:[WKWebView class]])`
But we will continue receiving the Apple warning on submission.

justinseanmartin: I'm thinking from a backwards compatibility perspective for projects that still rely on UIWebView. We shouldn't break the test framework from working on those projects while it is still allowed for app submissions.

Why are you linking KIF into your app for app store build? The KIF code generally shouldn't be present in Release artifacts, because it uses a bunch of private APIs. If you're using CocoaPods, you can use the pod `configuration` option (see [Podfile syntax](https://guides.cocoapods.org/syntax/podfile.html#pod)) to restrict it to only being linked in Debug builds for example.

dmatveichev:
@justinseanmartin please unblock this PR

justinseanmartin:
Subsumed the changes here in #1150.

PR #1126, opened 2019-10-10
Fixed acknowledgeSystemAlert on iOS 13.1

### LINKED ISSUE #1125: UIAutomationHelper.acknowledgeSystemAlertWithIndex Does Not Work In Xcode 11.1
`KIFUITestActor`'s `- (BOOL)acknowledgeSystemAlertWithIndex:(NSUInteger)index;`  no longer works when running on Xcode 11.1 and an iOS 13.1 simulator.

From my initial debugging, the alert `isKindOfClass` `nilElementClass` and it is never valid or visible.

ksuther:
We need to dig through the hierarchy manually rather than asking for the application's alert
Fixes #1125


justinseanmartin:
Thanks for the fix!

Issue #1127, opened 2019-10-15
-[UIAutomationHelper acknowledgeSystemAlertWithIndex] Does Not Work In Xcode 11.1


patmalt:
Related to #1125

The fix implemented in #1126, helped but does not completely solve the issue that system alerts cannot be dismissed.

The fix allows `![alert isKindOfClass:[self nilElementClass]]` to pass. However, ` [self _alertIsValidAndVisible:alert]` still fails. Also, `alert.buttons.count` is zero, so `isIndexInRange` is false.

It appears `[[alert valueForKeyPath:@"isVisible"] boolValue]` is throwing, in `_alertIsValidAndVisible`. Additionally, calling `[alert isVisible]` returns false, but does not throw.

https://github.com/kif-framework/KIF/blob/5735ec4fcf843bf2a0b7563f97e1fc8808aa4418/Classes/UIAutomationHelper.m#L149-L152

https://github.com/kif-framework/KIF/blob/5735ec4fcf843bf2a0b7563f97e1fc8808aa4418/Classes/UIAutomationHelper.m#L204-L207

justinseanmartin:
@patmalt - Please try out the latest master version of KIF and see if the latest PR merged fixes your issue, and reopen if the issue persists. Thanks!

I'll make a release with Xcode 11 fixes once things settle down and things are confirmed to be working as expected.

PR #1129, opened 2019-10-23
Fix bug when acknowledging system alerts in Xcode 11.1

briannewberry:
After upgrading to Xcode 11.1, KIF tests on our app were no longer able to acknowledge system notifications. After debugging it was found that the UIAlert is not always the first item in the list of elements. This change searches the list of elements for the UIAlertControllerView making it more resilient.

May help issue #1127


justinseanmartin:
Thanks for the contribution!

PR #1141, opened 2020-03-17
Scroll once when waiting for UITableViewCells

zradke:
Previously we would call `UITableView.scrollToRow(at:at:animated:)`
every runloop while we waited for the scroll to complete. This caused
multiple unnecessary calls to the delegate's
`scrollViewDidEndScrollingAnimation(_:)` method. This change resolves
the issue by moving the call to scroll to the row outside of the
`KIFTestActor.run(_:)` block.

### DIFF
--- a/Classes/KIFUITestActor.m
+++ b/Classes/KIFUITestActor.m
@@ -1251,10 +1251,13 @@ static BOOL KIFUITestActorAnimationsEnabled = YES;
         return KIFTestStepResultSuccess;
     }];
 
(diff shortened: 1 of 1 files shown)


zradke:
I also noted that the implementation of the UICollectionView's wait for cell method differs in it's implementation such that it doesn't do the incremental waiting and just waits for 0.5seconds. Not sure if I should update one or the other to match.

justinseanmartin:
I'd recommend leaving the collection view as is for now unless there is something specific that it is breaking and we can determine that the change is safe.

PR #1143, opened 2020-03-19
Widen `firstResponder` searches to all windows

zradke:
Previously searches for `UIWindow.firstResponder` were limited to the
`UIApplication.keyWindow`. In practice, it is very possible for users to
add additional valid windows that are not the key window and have first
responders. To accommodate those cases, this change broadens the search
for a first responder of an application to an array of first responders
found in `UIApplication.windowsWithKeyWindow`.

### DIFF
--- a/Classes/KIFUITestActor.m
+++ b/Classes/KIFUITestActor.m
@@ -220,7 +220,7 @@ static BOOL KIFUITestActorAnimationsEnabled = YES;
             [self waitForTimeInterval:maximumWaitingTimeInterval relativeToAnimationSpeed:YES];
         }
     } else {
-    
+
         // Wait for the view to stabilize and give them a chance to start animations before we wait for them.
         [self waitForTimeInterval:stabilizationTime relativeToAnimationSpeed:YES];
         maximumWaitingTimeInterval -= stabilizationTime;
@@ -468,7 +468,7 @@ static BOOL KIFUITestActorAnimationsEnabled = YES;
     [text enumerateSubstringsInRange:NSMakeRange(0, text.length)
                              options:NSStringEnumerationByComposedCharacterSequences
                           usingBlock: ^(NSString *characterString,NSRange substringRange,NSRange enclosingRange,BOOL * stop)
-    {
+     {
         if (![KIFTypist enterCharacter:characterString]) {
             NSLog(@"KIF: Unable to find keyboard key for %@. Will attempt to insert manually.", characterString);
 
@@ -480,7 +480,7 @@ static BOOL KIFUITestActorAnimationsEnabled = YES;
             } else {
                 [fallbackViews addObjectsFromArray:[[UIApplication sharedApplication] firstResponders]];
             }
-
+            
             for (id fallback in [fallbackViews copy]) {
                 if (![fallback isKindOfClass:[UITextField class]] &&
                     ![fallback isKindOfClass:[UITextView class]] &&
@@ -998,7 +998,7 @@ static BOOL KIFUITestActorAnimationsEnabled = YES;
 
     // Tap the desired photo in the grid
     // TODO: This currently only works for the first page of photos. It should scroll appropriately at some point.
-     UIAccessibilityElement *headerElt = [[UIApplication sharedApplication] accessibilityElementMatchingBlock:^(UIAccessibilityElement *element) {
+    UIAccessibilityElement *headerElt = [[UIApplication sharedApplication] accessibilityElementMatchingBlock:^(UIAccessibilityElement *element) {
         return [NSStringFromClass(element.class) isEqual:@"UINavigationItemButtonView"];
     }];
     UIView* headerView = [UIAccessibilityElement viewContainingAccessibilityElement:headerElt];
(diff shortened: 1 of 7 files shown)


Classes/KIFUITestActor.m
```diff
@@ -471,21 +471,27 @@ - (void)enterTextIntoCurrentFirstResponder:(NSString *)text fallbackView:(UIView
...
 
-            if ([fallback isKindOfClass:[UITextField class]] || [fallback isKindOfClass:[UITextView class]] || [fallback isKindOfClass:[UISearchBar class]]) {
-                NSLog(@"KIF: Unable to find keyboard key for %@. Inserting manually.", characterString);
-                [(UITextField *)fallback setText:[[(UITextField *)fallback text] stringByAppendingString:characterString]];
-            } else {
-                [self failWithError:[NSError KIFErrorWithFormat:@"Failed to find key for character \"%@\"", characterString] stopTest:YES];
+            for (UIView *fallback in fallbackViews) {
+                if ([fallback isKindOfClass:[UITextField class]] || [fallback isKindOfClass:[UITextView class]] || [fallback isKindOfClass:[UISearchBar class]]) {
+                    NSLog(@"KIF: Unable to find keyboard key for %@. Inserting manually.", characterString);
+                    [(UITextField *)fallback setText:[[(UITextField *)fallback text] stringByAppendingString:characterString]];
```
justinseanmartin: I wonder if we should print a warning if there is more than one match, to make the surface the somewhat arbitrary decision being made here.

Classes/KIFUITestActor.m
```diff
@@ -551,9 +557,10 @@ - (void)expectView:(UIView *)view toContainText:(NSString *)expectedResult
 - (void)clearTextFromFirstResponder
 {
     @autoreleasepool {
-        UIView *firstResponder = (id)[[[UIApplication sharedApplication] keyWindow] firstResponder];
-        if ([firstResponder isKindOfClass:[UIView class]]) {
-            [self clearTextFromElement:(UIAccessibilityElement *)firstResponder inView:firstResponder];
+        for (UIResponder *firstResponder in [[UIApplication sharedApplication] firstResponders]) {
+            if ([firstResponder isKindOfClass:[UIView class]]) {
+                [self clearTextFromElement:(UIAccessibilityElement *)firstResponder inView:(UIView *)firstResponder];
```
justinseanmartin: We probably should only do this from one element and not multiple.

Classes/KIFUITestActor.m
```diff
@@ -1182,13 +1189,24 @@ - (void)scrollAccessibilityElement:(UIAccessibilityElement *)element inView:(UIV
...
-        if ([firstResponder isKindOfClass:NSClassFromString(@"UISearchBarTextField")]) {
-            do {
-                firstResponder = [(UIView *)firstResponder superview];
-            } while (firstResponder && ![firstResponder isKindOfClass:[UISearchBar class]]);
+        BOOL didMatch = NO;
+        UIResponder *foundResponder = nil;
+        for (UIResponder  *firstResponder in [[UIApplication sharedApplication] firstResponders]) {
+            foundResponder = firstResponder;
+
+            if ([foundResponder isKindOfClass:NSClassFromString(@"UISearchBarTextField")]) {
```
justinseanmartin: :eyeroll: at this special case being here, but not for the other method w/ traits matching below... (no action for you, just noticing this while reviewing)

Classes/KIFUITestActor.m
```diff
@@ -1182,13 +1189,24 @@ - (void)scrollAccessibilityElement:(UIAccessibilityElement *)element inView:(UIV
...
+            }
+
+            if (foundResponder.accessibilityLabel == label || [foundResponder.accessibilityLabel isEqualToString:label]) {
+                didMatch = YES;
+                break;
+            }
         }
-        KIFTestWaitCondition([[firstResponder accessibilityLabel] isEqualToString:label], error, @"Expected accessibility label for first responder to be '%@', got '%@'", label, [firstResponder accessibilityLabel]);
+
+        KIFTestWaitCondition(didMatch, error, @"Expected accessibility label for first responder to be '%@', got '%@'", label, [foundResponder accessibilityLabel]);
```
justinseanmartin: Maybe the assertions should use dump the whole array of elements (or maybe at least the count of elements checked)? (same applies below)

IdentifierTests/KIFUITestActor-IdentifierTests.m
```diff
@@ -158,25 +159,28 @@ - (void)setValue:(float)value forSliderWithAccessibilityIdentifier:(NSString *)a
...
-		}
-		else
-		{
-			[self failWithError:[NSError KIFErrorWithFormat:@"First responder does not conform to UIAccessibilityIdentification %@",  NSStringFromClass([firstResponder class])] stopTest:YES];
-			
-		}
-		KIFTestWaitCondition([[firstResponderIdentification accessibilityIdentifier] isEqualToString:accessibilityIdentifier],
+        BOOL didMatch = NO;
+        NSString *foundIdentifier = nil;
+        for (UIResponder __strong *firstResponder in [[UIApplication sharedApplication] firstResponders]) {
```
justinseanmartin: Why. is the `__strong` needed here?

Classes/KIFUITestActor.m
zradke: Do you mean that I should revert this portion so it continues to only check `UIApplication.keyWindow`?

IdentifierTests/KIFUITestActor-IdentifierTests.m
zradke: It lets me reassign to `firstResponder` in the `for` loop without having to declare a temporary local variable.

Classes/KIFUITestActor.m
zradke: Yeah I noticed a lot of inconsistencies around `accessibilityLabel` checking. Might warrant a deeper follow up to unify that logic.

justinseanmartin: In general, `KIFUITestActor` is the old way of doing things and should eventually be deprecated. It probably isn't worth the time, as everything is already unified in `KIFUIViewTestActor` (new way).

Classes/KIFUITestActor.m
justinseanmartin: I guess what I mean is to bail from the loop after clearing the text from some first responder. Probably would want to do the same as above and output a warning if there are multiple matching fields and we're only clearing the contents of one of them.

IdentifierTests/KIFUITestActor-IdentifierTests.m
```diff
@@ -158,25 +159,28 @@ - (void)setValue:(float)value forSliderWithAccessibilityIdentifier:(NSString *)a
...
+                foundIdentifier = [(UIResponder<UIAccessibilityIdentification> *)firstResponder accessibilityIdentifier];
+            }
+
+            if ([foundIdentifier isEqualToString:accessibilityIdentifier]) {
+                didMatch = YES;
+                break;
+            }
+        }
+
+		KIFTestWaitCondition(didMatch,
```
justinseanmartin: NIT: indentation seems off

IdentifierTests/KIFUITestActor-IdentifierTests.m
justinseanmartin: I'd probably still opt to use the local variable to make the logic more straightforward, but don't feel strongly on this.

IdentifierTests/KIFUITestActor-IdentifierTests.m
zradke: Hmm looks like there's a mix of spaces & tabs in this file, so it looks fine with my config but wrong for github's. I'll fix all in file to convert them all to spaces. It'll dirty up the commit so I'd advise ignoring whitespace afterward.

Classes/KIFUITestActor.m
```diff
@@ -468,24 +468,25 @@ - (void)enterTextIntoCurrentFirstResponder:(NSString *)text fallbackView:(UIView
...
+
             // Attempt to cheat if we couldn't find the character
-            UIView * fallback = fallbackView;
-            if (!fallback) {
-                UIResponder *firstResponder = [[[UIApplication sharedApplication] keyWindow] firstResponder];
+            NSArray *fallbackViews = fallbackView ? @[fallbackView] : [[UIApplication sharedApplication] firstResponders];
+
+            for (id fallback in fallbackViews) {
+                if ([fallback isKindOfClass:[UITextField class]] || [fallback isKindOfClass:[UITextView class]] || [fallback isKindOfClass:[UISearchBar class]]) {
+                    if (fallbackViews.count > 1) {
```
justinseanmartin: I was thinking of gathering the set of first responders that match the class type check and then output the log message only if there is more than one match. Just concerned that this might be too noisy if it is an expected common scenario.

IdentifierTests/KIFUITestActor-IdentifierTests.m
justinseanmartin: Thanks for taking that up!

justinseanmartin:
LGTM. Take a look at my one latest comment and LMK what you think. I'll merge this as-is if you don't think it is worth addressing.

Classes/KIFUITestActor.m
zradke: Fixed!

Issue #1130, opened 2019-10-24
KIF for swift package Xcode 11.1?


vitinholiro:

justinseanmartin:
We don't currently have a SPM manifest for any versions of KIF. If you'd like to make one that works for folks, I'd be happy to merge the pull request into the repo.

kenji21:
Hi,

Initial pull request sent : https://github.com/kif-framework/KIF/pull/1147

kenji21:
New PR replacing previous one : https://github.com/kif-framework/KIF/pull/1193

heyzooi:
@dostrander I'm wondering why you guys had to revert the support for SPM. We are using 3.7.14 but the support disappeared for 3.7.15

dostrander:
@heyzooi The issue with this pull request is it had changes seemingly unrelated to SPM that broke cocoapods. Somehow while releasing 3.7.14 it still published however broke when pod installing. 3.7.15 fixed that by reverting to allow for cocoapods to continue working.

I put up a PR to verify that cocoapods stays in tact while merging pull requests.

I am planning on getting this change back into KIF as well as some improvements to our testing so we have more safeguards in place by the end of this week or at the latest next week if you could hold off on 3.7.14

PR #1147, opened 2020-04-23
#1130 Initial swift package manager support

kenji21:
I had to remove pre-compiled header files (.pch) because there is no such thing in swift package manager. It leads to a lot of "#import" changes.

Hard to find how to link with IOKit (tried many "hacks"), it was really simple in fact (in Package.swift):

```
            linkerSettings: [.linkedFramework("IOKit")]),
```

Another issue was with Tests files that are including some "private" headers (in Package.swift):

```
           cSettings: [.headerSearchPath("../Sources/KIF/")]), // allow to look a "private" headers
```

Also had to move files around to be on the SPM standard paths (specially for public headers in include directory)

Built on Xcode 11.4.1 / iPhone 11 Pro (iOS 13.4.1) simulator with both "old xcodeproj" and Package.swift opened using Xcode (which of course only show the KIF/KIFTests target)

Used documentations:

https://github.com/apple/swift-package-manager/blob/master/Documentation/
https://github.com/apple/swift-package-manager/blob/master/Documentation/Usage.md
https://github.com/apple/swift-package-manager/blob/master/Documentation/PackageDescription.md


justinseanmartin:
First of all, huge thanks for taking this on! Overall the change seems great directionally, and I don't really have issue with removing the bucketing of files between `Classes` and `Additions`.

Just reading "Creating C Language Targets" here to try and understand some of the requirements herer:
https://github.com/apple/swift-package-manager/blob/master/Documentation/Usage.md

It seems like we should be able to flatten the directory structure a bit. Aren't the two `KIF` folders under `Sources` superfluous? In other words, is it possible to move the files from `Sources/KIF` to live under `Sources` and move the public headers from `Sources/KIF/include/KIF` to live under `Sources/include`?

justinseanmartin:
We should add a CI job that validates the SPM package can be compiled properly and linked into a test target and used to run a simple integration test case. This will ensure that we won't regress this functionality going forward. 

Would this be something you'd mind looking into? If not, there is definitely some risk this breaks accidentally with a future contribution. That said, our current CI configuration is also woefully out-of-date and I don't have the time right now to commit to updating it, so I understand it seeming a bit hypocritical. It could potentially be a special CI configuration only used for this SPM integration test, so we don't have to get all the integration tests passing on Xcode 11.4.

kenji21:
Ok, no problem to move all sources files to `Sources/KIF`, maybe I should also move Tests sources files to `Tests/KIFTests`

For the CI, does using `./Documentation/Examples/Testable Swift/Testable Swift.xcodeproj` could be a good idea (adding the dependency via its path `../../../KIF`) ?

justinseanmartin:
Regarding the `Sources/KIIF`, what I meant was, why not move everything to `Sources/*` and `Sources/includes/*`? There's definitely a chance I'm missing something context though, as I haven't worked with SPM very much.

The `Testable Swift.xcodeproj` already is hardcoded to link against the KIF.framework produced by building KIF.xcodeproj. I think we'd want another sample project (could probably be a copy of that one to start, but also could be a File > New Project) with a SPM package reference to integrate KIF into that sample.

kenji21:
Not moving all sources was to make less modifications as possible to the pbxproj

Just duplicated the Testable Swift target (and test target) to make it use KIF via SPM (using a file:///$(SRCROOT) url)

justinseanmartin:
Sorry for the delay, I'll try to get back to this later this week, just incredibly crunched for time at the moment.

justinseanmartin:
I was just reading up on this a bit and I like the suggestion here of using symlinks, rather than breaking colocation of implementation and interfaces:

https://medium.com/@joesusnick/swift-package-manager-with-a-mixed-swift-and-objective-c-project-part-2-2-e71dad234e6

Do you think this is reasonable @kenji21? Sorry for the long silence on my part, I've had a lot going on.

justinseanmartin:
Overall though, this last iteration looks much better than reorganizing all the source code in the project just to appease SPM. Thanks for iterating on this!

kenji21:
can't get CI to work with this, got this while opening project with Xcode 11.5:

```
The targets “Testable SwiftViaPM” and “Testable SwiftViaPMTests” contains source code developed with Swift 3.x. This version of Xcode does not support building or migrating Swift 3.x targets.

Use Xcode 10.1 to migrate the code to Swift 4.
```

justinseanmartin:
Rather than putting the target in the existing `Documentation/Examples/Testable Swift` folder, would it be helpful making a new example under `Documentation/Examples/SwiftPM`? That way, we can change all the build settings to be updated for Swift 5.2. 

We'd need to add a new Travis CI configuration to run on Xcode 11. I can help setting that up if you can get the SwiftPM project split out from the existing "Testable Swift" project.

I appreciate you continuing to work on this!

kenji21:
ok... code climate found that I duplicated the sample projet...

kenji21:
and CI is fixed \o/ (but test of Testable SwiftPM were not run)

justinseanmartin:
@dostrander ^ when you get a chance. Happy to sync up offline.

dostrander:
Will take a look at this early next week @justinseanmartin @kenji21 so we can get this in ASAP. Sorry for the delay @kenji21

bdrelling:
Hey, all! 

Firstly, thanks for the hard work on this! Second, anything I can do to help push this forward so we can migrate over to SPM?

Thanks!

smpeleato:
Hi everyone!

Thanks for all the work on Swift Package Manager support! I also wanted to ask if there was anything I can do to help, as I'm also eager to include KIF via SPM.

Thanks,
Sebastian

edulpn:
I double this, since Carthage started having issues with Xcode 12 I'm looking into migrating to SPM and can help with anything that is necessary to make this work!

fruitcoder:
Same here! Can't wait for spm 🙂

kenji21:
Closing, will continue at https://github.com/kif-framework/KIF/pull/1193

PR #1150, opened 2020-06-12
Support WKWebview (UIWebView support deprecated in iOS 12)

justinseanmartin:
Adding back support for UIWebView. Picking up landing the changes from #1121, never heard back from the original author.

CC: @dmatveichev @ggdiez @damirstuhec


harleyjcooper:
LGTM! Seems straightforward.

dmatveichev:
@justinseanmartin thank you! But, I don't think there is a need to support UIWebView. Starting Dec 2020 if you use UIWebView you will not be able to submit your app in AppStore. So everyone switching to WKWebView now.

PR #1151, opened 2020-06-24
Fix compilation issue for Xcode 12

danl3v:
Add support for Xcode 12. I encountered compilation failures when building this dependency with Carthage.

### DIFF
--- a/Additions/XCTestCase-KIFAdditions.m
+++ b/Additions/XCTestCase-KIFAdditions.m
@@ -34,7 +34,18 @@ static inline void Swizzle(Class c, SEL orig, SEL new)
 {
     self.continueAfterFailure = YES;
 
-    [self recordFailureWithDescription:exception.description inFile:exception.userInfo[@"FilenameKey"] atLine:[exception.userInfo[@"LineNumberKey"] unsignedIntegerValue] expected:NO];
+    NSString *filePath = exception.userInfo[@"FilenameKey"];
+    NSInteger lineNumber = [exception.userInfo[@"LineNumberKey"] unsignedIntegerValue];
+    XCTSourceCodeLocation *location = [[XCTSourceCodeLocation alloc] initWithFilePath:filePath
+                                                                           lineNumber:lineNumber];
+    XCTSourceCodeContext *context = [[XCTSourceCodeContext alloc] initWithLocation:location];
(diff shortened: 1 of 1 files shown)


Additions/XCTestCase-KIFAdditions.m
```diff
@@ -34,7 +34,18 @@ - (void)failWithException:(NSException *)exception stopTest:(BOOL)stop
 {
     self.continueAfterFailure = YES;
 
-    [self recordFailureWithDescription:exception.description inFile:exception.userInfo[@"FilenameKey"] atLine:[exception.userInfo[@"LineNumberKey"] unsignedIntegerValue] expected:NO];
+    NSString *filePath = exception.userInfo[@"FilenameKey"];
```
justinseanmartin, line 37: Would you mind wrapping this and below usages of `XCTIssue` in a `#ifdef __IPHONE_14_0` and put the original implementation in the `#else`? That way this is backwards compatible.

Additions/XCTestCase-KIFAdditions.m
```diff
@@ -34,7 +34,18 @@ - (void)failWithException:(NSException *)exception stopTest:(BOOL)stop
...
+    NSInteger lineNumber = [exception.userInfo[@"LineNumberKey"] unsignedIntegerValue];
+    XCTSourceCodeLocation *location = [[XCTSourceCodeLocation alloc] initWithFilePath:filePath
+                                                                           lineNumber:lineNumber];
+    XCTSourceCodeContext *context = [[XCTSourceCodeContext alloc] initWithLocation:location];
+    XCTIssue *issue = [[XCTIssue alloc] initWithType:XCTIssueTypeAssertionFailure
+                                  compactDescription:exception.description
+                                 detailedDescription:nil
+                                   sourceCodeContext:context
+                                     associatedError:nil
+                                         attachments:@[]];
```
justinseanmartin, line 47: Out of scope - It would potentially make sense to attach the generated test failure screenshot and the view hierarchy as attachments at some point. Not in this PR though.

danl3v:
Will do! Thanks for the review and feedback

Additions/XCTestCase-KIFAdditions.m
danl3v: That would be fancy!

PR #1154, opened 2020-06-30
Add CI configuration for iOS 12 

dostrander:
Also fixes issues that arose with ios 12

### DIFF
--- a/KIF Tests/AccessibilityIdentifierTests.m
+++ b/KIF Tests/AccessibilityIdentifierTests.m
@@ -47,12 +47,14 @@
 
 - (void)testLongPressingViewWithAccessibilityIdentifier
 {
+  [tester tapViewWithAccessibilityIdentifier:@"idGreeting"];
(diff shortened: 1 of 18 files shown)


KIF Tests/AccessibilityIdentifierTests.m
```diff
@@ -47,12 +47,14 @@ - (void)testWaitingForAbscenceOfViewWithAccessibilityIdentifier
 
 - (void)testLongPressingViewWithAccessibilityIdentifier
 {
+  [tester tapViewWithAccessibilityIdentifier:@"idGreeting"];
```
erudel, line 50: I wonder if it makes sense to add a `showMenu` text action (possibly with a version check if backwards compatibility is an issue) that executes the tap+long press.

dostrander: Would you add an issue for an improvement to add that and we will get to that when we can.

dostrander:
Closed: Updating using https://github.com/kif-framework/KIF/pull/1168

Issue #1134, opened 2019-12-10
In iPhone 11 Simulator, iOS 13.2.2 , NavigationBarBackItem has a new class '_UIModernBarButton'.


edisongz:
_UIModernBarButton's  userInteractionEnabled property has a default value NO, this can be fixed by append a class type for `- (void)isNavigationItemView;`

justinseanmartin:
Cool, thanks for reporting. Would you mind submitting a patch with the fix, as you're currently set up to test and ensure that it is working as expected. Thanks!

erudel:
I put up a PR to fix this issue: https://github.com/kif-framework/KIF/pull/1158

I verified some of the failing unit tests are now passing on Xcode 11 / iOS 13.4

PR #1158, opened 2020-07-07
Fix navigation back button tap (issue #1134)

erudel:
For more recent iOS versions the internal view hierarchy of navigation bar buttons has changed. A lot of unit tests are broken because it is not possible to tap on the back button anymore.

This PR adds a check for marking the new button as a navigation item view.


dostrander:
Did the same thing here: https://github.com/kif-framework/KIF/pull/1154/files but let's take yours to get in faster for folks who need it.

erudel:
Thanks @dostrander, I only saw your PR after posting mine. I've been seeing other issues when running the main test suite and would like to start digging a bit more whenever I have a chance.

PR #1170, opened 2020-08-27
Wait for animations to complete after performing a scroll action

justinseanmartin:
We're seeing an issue where we're unable to tap on an element right after performing a `-[KIFUIViewTestActor scrollByFractionOfSizeHorizontal:vertical:]` action. It appears to help ensure that the view has stabilized before attempting to interact with that UI element.


segiddins:
Is there a test we can add for this fix?

justinseanmartin:
Declined in favor of #1210, which has a test

PR #1179, opened 2020-09-15
Adds a visualizer to KIF

dostrander:
Adds a touch visualizer to see where touches are being sent to. Currently it has a default look and feel. We can have a configuration object that can get set if we feel in the future we want customizability over the visualization of touches.

This is an explicit opt-in feature similar to printing the view hierarchy.

### DIFF
--- a/Visualizer/KIFEventVisualizer.h
+++ b/Visualizer/KIFEventVisualizer.h
@@ -1,0 +1,14 @@
+#import <Foundation/Foundation.h>
+
+@interface KIFEventVisualizer : NSObject
+
++ (nonnull instancetype)sharedVisualizer;
+
+- (void)visualizeEvent:(nonnull UIEvent *)event;
+
+// Unavailable.
+- (nonnull instancetype)init NS_UNAVAILABLE;
++ (nonnull instancetype)new NS_UNAVAILABLE;
+
+
+@end
--- a/Additions/UIApplication-KIFAdditions.h
+++ b/Additions/UIApplication-KIFAdditions.h
@@ -118,6 +118,12 @@ CF_EXPORT SInt32 KIFRunLoopRunInModeRelativeToAnimationSpeed(CFStringRef mode, C
  */
 + (void)stopMockingOpenURL;
 
+/*
+ Sends event to visualizer and application.
+ */
+- (void)kif_sendEvent:(UIEvent *)event;
+
+
 @end
 
 @interface UIApplication (Private)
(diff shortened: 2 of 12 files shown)


justinseanmartin (approved):
Super stoked about this landing!

Visualizer/KIFEventVisualizer.h
```diff
@@ -0,0 +1,9 @@
+#import <Foundation/Foundation.h>
+
+@interface KIFEventVisualizer : NSObject
+
++ (instancetype)sharedVisualizer;
```
harleyjcooper, line 5: Do we want to set `- (instancetype) init` to `NS_UNAVAILABLE`? Or would anyone else be calling the `init` method?

harleyjcooper (approved):
This is really cool! One small comment, but it otherwise looks awesome.

Visualizer/KIFEventVisualizer.h
dostrander: Good callout, since we are creating our own window overlay we probably want only 1 being created.

dostrander: Had to make an `initPrivate` method because since it's marked unavailable using `init` internally also breaks

Visualizer/KIFEventVisualizer.h
```diff
@@ -0,0 +1,14 @@
+#import <Foundation/Foundation.h>
+
+@interface KIFEventVisualizer : NSObject
+
++ (nonnull instancetype)sharedVisualizer;
+
+- (void)visualizeEvent:(nonnull UIEvent *)event;
+
+// Unavaiable.
```
harleyjcooper, line 9: *unavailable

Visualizer/KIFEventVisualizer.h
harleyjcooper: Looks good!

jgavris:
I don't believe the CocoaPods podspec includes these new files, we had been running KIF at HEAD for a while and had a build failure introduced with this PR. I see that the CI only tests Carthage. Do most people use Carthage these days?

dostrander:
@jgavris I'll fix that this morning, sorry about that. I don't think most use Carthage. Most probably aren't on HEAD though.

I'll also fix up the project so we catch this in CI (or earlier).

jgavris:
@dostrander That would be excellent! Thanks. For our internal pods, we just have a really simple example project that imports the pod and checks that it builds.

jgavris:
At the same time, for third-party libraries, we are starting to recommend not building from source, as those iOS build times grow and grow. So Carthage with something like Rome, or Cocoapods with the [binary plugin](https://github.com/leavez/cocoapods-binary), but that also has to lean on good CI in the external libraries.

Issue #1181, opened 2020-09-16
dismissPopover does not work in iOS 14


dostrander:
In iOS 14 there is no `UIDimmingView` that can be detected to dismiss a popover.

There is a `_UITouchFallbackView` which we can use to tap and dismiss a popover view.

dostrander:
Fixed with #1182

PR #1182, opened 2020-09-16
Fix Issue #1181 - `dismissPopover` does not work on iOS 14

dostrander:
`UIDimmingView` looks to be replaced with `_UITouchFallbackView` in iOS 14. This PR allows us to fallback to `_UITouchFallbackView` when `UIDImmingView` isn't found.

### DIFF
--- a/Classes/KIFUITestActor.m
+++ b/Classes/KIFUITestActor.m
@@ -960,8 +960,12 @@ static BOOL KIFUITestActorAnimationsEnabled = YES;
     if (!window) {
         [self failWithError:[NSError KIFErrorWithFormat:@"Failed to find any dimming views in the application"] stopTest:YES];
     }
-    UIView *dimmingView = [[window subviewsWithClassNamePrefix:@"UIDimmingView"] lastObject];
-    [dimmingView tapAtPoint:CGPointMake(50.0f, 50.0f)];
+    UIView *touchView = [[window subviewsWithClassNamePrefix:@"UIDimmingView"] lastObject];
+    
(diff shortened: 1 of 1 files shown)


Classes/KIFUITestActor.m
```diff
@@ -1164,7 +1168,7 @@ - (void)pullToRefreshAccessibilityElement:(UIAccessibilityElement *)element inVi
     // Can handle only the touchable space.
     CGRect elementFrame = [viewToSwipe convertRect:viewToSwipe.bounds toView:[UIApplication sharedApplication].keyWindow.rootViewController.view];
     CGPoint swipeStart = CGPointCenteredInRect(elementFrame);
-    CGPoint swipeDisplacement = CGPointMake(CGRectGetMidX(elementFrame), CGRectGetMaxY(elementFrame));
+    CGPoint swipeDisplacement = CGPointMake(CGRectGetMidX(elementFrame) - swipeStart.x, CGRectGetMaxY(elementFrame) - swipeStart.y);
```
harleyjcooper, line 1171: What change led to this bit? Why has the displacement changed?

harleyjcooper (approved):
LGTM other than one question.

Classes/KIFUITestActor.m
dostrander: Oops didn't mean to change this.

The reason why i even put this in here at all was when I put in the visualizer i noticed the the pull to refresh went diagonally this fixes that. Not an issue so going to leave as is for now but just was testing things out and forgot to remove it

PR #1185, opened 2020-09-21
Fix pull to refresh displacement to not go outside of the touchable view and drag straight down.

dostrander:
Before the pul to refresh mechanism went diagonal and would go far off screen because we were using the views MaxY as a displacement rather than the difference between the start of the swipe and the end of the view.


harleyjcooper (approved):
Thanks for doing this!

justinseanmartin (approved):
Overall LGTM, but probably shouldn't merge until red builds get looked at.

mikelupo:
Can’t recall why we had to make this not drag straight down. Anomaly with older devices ( real not simulators) comes to mind. Please don’t change that

Issue #1183, opened 2020-09-16
Selecting time or date in date picker does not work in iOS14


dostrander:
In iOS 14 they have changed the UI of the date picker. KIF needs to be updated to reflect this change in iOS14.

When selecting date picker input:
![Simulator Screen Shot - iPad (8th generation) - 2020-09-16 at 12 46 16](https://user-images.githubusercontent.com/823322/93367710-ddaa8700-f81a-11ea-91e7-7bc69ff92257.png)

Once you click the data/time!
![Simulator Screen Shot - iPad (8th generation) - 2020-09-16 at 12 46 19](https://user-images.githubusercontent.com/823322/93367739-e9964900-f81a-11ea-807f-a123683509c4.png)

dostrander:
Note this is only when developers do NOT set `preferredDatePickerStyle` to `UIDatePickerStyleWheels`. If it's set to `UIDatePickerStyleWheels` the original implementation still works.

PR #1187, opened 2020-09-23
Fixes Issue #1183 where date pickers don't work in iOS 14

dostrander:
With iOS 14 there is a new calendar style date picker which breaks how we use the date picker that we assumed would be wheels. 
This PR deprecates `selectDatePickerValue:` and associated methods to have a more clear API on what dates should be selected regardless of style of date picker (calendar vs wheel). We currently have naive backwards compatibility with these methods but no guarantee they work in every locale, language, or situation. These deprecated methods should be moved over to the new API as soon as possible and have the possibility of being removed in the future.


Test Host/PickerController.m
```diff
@@ -44,6 +48,9 @@ - (void)viewDidLoad
 
     datePicker = [[UIDatePicker alloc] initWithFrame:CGRectMake(30, 215, 260, 35)];
     datePicker.datePickerMode = UIDatePickerModeDate;
+    if(@available(iOS 13.4, *)) {
+        datePicker.preferredDatePickerStyle = UIDatePickerStyleWheels;
```
justinseanmartin, line 52: Might help to explicitly rename `datePicker` to `wheelDatePicker` or something to make the distinction more clear.

Test Host/PickerController.m
```diff
@@ -99,50 +115,100 @@ - (void)viewDidLoad
     countdownSelectionTextField.returnKeyType = UIReturnKeyDone;
     countdownSelectionTextField.inputView = countdownPicker;
     countdownSelectionTextField.accessibilityLabel = @"Countdown Selection";
+    
+    self.dateCalendarPicker = [[UIDatePicker alloc] init];
```
justinseanmartin, line 119: I assume these are using `preferredDatePickerStyle` of automatic? Should we be more explicit here and choose `UIDatePickerStyleCompact`? Should we be handling `UIDatePickerStyleInline` as well, or considering that out of scope for now?

justinseanmartin: Also, should these fields only be exposed and tested for iOS 14+? Looks like compact mode is supported in 13.4+ from the docs.

Test Host/PickerController.m
```diff
@@ -99,50 +115,100 @@ - (void)viewDidLoad
...
+    self.dateCalendarPicker.datePickerMode = UIDatePickerModeDate;
+    [self.dateCalendarPicker addTarget:self action:@selector(dateCalendarPickerChanged:) forControlEvents:UIControlEventValueChanged];
+    [self.dateCalendarPicker setLocale:[[NSLocale alloc] initWithLocaleIdentifier:@"en_US"]];
+    
+    self.datePickerCalendarTextField.placeholder = NSLocalizedString(@"Date Calendar Selection", nil);
+    self.datePickerCalendarTextField.returnKeyType = UIReturnKeyDone;
+    self.datePickerCalendarTextField.inputView = self.dateCalendarPicker;
+    self.datePickerCalendarTextField.accessibilityLabel = @"Date Calendar Selection";
```
justinseanmartin, line 127: AFAICT - Not seeing any tests that are interacting with `Date Calendar Selection` or `Date Time  Calendar Selection`? Totally might be missing it.

Test Host/PickerController.m
```diff
@@ -99,50 +115,100 @@ - (void)viewDidLoad
...
+    
+    self.dateTimeCalendarPicker = [[UIDatePicker alloc] init];
+    self.dateTimeCalendarPicker.datePickerMode = UIDatePickerModeDateAndTime;
+    [self.dateTimeCalendarPicker addTarget:self action:@selector(dateTimeCalendarPickerChanged:) forControlEvents:UIControlEventValueChanged];
+    [self.dateTimeCalendarPicker setLocale:[[NSLocale alloc] initWithLocaleIdentifier:@"en_US"]];
+    
+    self.dateTimePickerCalendarTextField.placeholder = NSLocalizedString(@"Date Time Calendar Selection", nil);
+    self.dateTimePickerCalendarTextField.returnKeyType = UIReturnKeyDone;
+    self.dateTimePickerCalendarTextField.inputView = self.dateTimeCalendarPicker;
+    self.dateTimePickerCalendarTextField.accessibilityLabel = @"Date Time  Calendar Selection";
```
justinseanmartin, line 137: NIT: Extra space in the middle of the name here.

Classes/KIFUITestActor.m
```diff
@@ -676,18 +682,167 @@ - (void)selectPickerViewRowWithTitle:(NSString *)title inComponent:(NSInteger)co
...
 
-- (void)selectDatePickerValue:(NSArray *)datePickerColumnValues fromPicker:(UIPickerView *)picker withSearchOrder:(KIFPickerSearchOrder)searchOrder
+- (void)selectDatePickerValue:(NSArray *)datePickerColumnValues fromPicker:(UIDatePicker *)datePicker withSearchOrder:(KIFPickerSearchOrder)searchOrder
 {
-    [self selectPickerValue:datePickerColumnValues fromPicker:picker pickerType:KIFUIDatePicker withSearchOrder:searchOrder];
+    NSDateComponents *dateComponents = [[NSDateComponents alloc] init];
+    NSArray *dateOrder = [self.class dateOrderForDatePicker:datePicker];
+    
+    // Backwards compatibility while migrating away from this deprecated API.
+    if(datePicker.datePickerMode == UIDatePickerModeDate) {
```
justinseanmartin, line 785: Should this logic take `datePickerStyle` into consideration for how to behave? It looks like wheel mode isn't gone in iOS 14, it just isn't the default.

dostrander: It doesn't care with the new way we do date picking, it works with any picker mode

dostrander: So my thinking is just move all types over to this

Test Host/PickerController.m
dostrander: Yep i have them in 13.4+. Inline isn't any different of styles just is that is already tested so i was thinking we can just  test the compact and wheel types

Classes/KIFUITestActor.m
```diff
@@ -676,18 +682,167 @@ - (void)selectPickerViewRowWithTitle:(NSString *)title inComponent:(NSInteger)co
     [self selectPickerViewRowWithTitle:title inComponent:component fromPicker:nil withSearchOrder:searchOrder];
 }
 
+- (void)selectCountdownDatePickerWithAccesibilityLabel:(NSString *)accesibilityLabel inputHours:(NSInteger)hours minutes:(NSInteger)minutes
```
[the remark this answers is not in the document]
dostrander: Moved these and the tests to viewtester only.

Classes/KIFUITestActor.m
```diff
@@ -676,18 +682,167 @@ - (void)selectPickerViewRowWithTitle:(NSString *)title inComponent:(NSInteger)co
...
+    [self selectDatePicker:datePicker inputDate:date];
+}
+
+- (void)selectInputDatePickerWithDate:(NSDate *)date
+{
+    UIDatePicker *datePicker = [[[[UIApplication sharedApplication] datePickerWindow] subviewsWithClassNamePrefix:@"UIDatePicker"] firstObject];
+    [self selectDatePicker:datePicker inputDate:date];
+}
+
+- (void)selectDatePicker:(UIDatePicker *)datePicker inputDate:(NSDate *)date
```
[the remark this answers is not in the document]
dostrander: done

Test Host/PickerController.m
dostrander: They are in the tests

dostrander:
@justinseanmartin Addressed comments

Classes/KIFUITestActor.m
```diff
@@ -43,6 +44,12 @@ @interface KIFUITestActor ()
 
 static BOOL KIFUITestActorAnimationsEnabled = YES;
 
+@interface UIDatePicker()
```
justinseanmartin, line 47: No longer needed here?

Additions/PrivateAPI/UIDatePicker+KIFPrivateAPI.h
```diff
@@ -0,0 +1,10 @@
+#import <UIKit/UIKit.h>
+
+// Private API's used for UIDatePicker.
+@interface UIDatePicker (KIFPrivateAPI)
+
+// After updating the value this is called to inform
+// the observers that the value has changed.
+- (void)_emitValueChanged;
```
justinseanmartin, line 8: NIT: Might be better to put this under some other folder to make it clear that it is exposing/relying on underlying APIs in Apple's SDK. Maybe a new root folder like `ApplePrivateAPIs`? I think there is some stuff within `IOHIDEvent+KIF.m` that would also potentially make sense moving there as well.

Additions/UIDatePicker+KIFAdditions.m
```diff
@@ -0,0 +1,21 @@
...
+#import "UIDatePicker+KIFPrivateAPI.h"
+
+@implementation UIDatePicker (KIFAdditions)
+
+- (void)selectDate:(NSDate *)date
+{
+    NSAssert(self.datePickerMode != UIDatePickerModeCountDownTimer, @"Date picker was not in expected date picking mode mode. Instead got countdown timer.");
+    self.date = date;
+    [self _emitValueChanged];
+
```
justinseanmartin, line 11: NIT: whitespace (human SpaceCommander...)

Classes/KIFUITestActor.m
```diff
@@ -676,18 +683,126 @@ - (void)selectPickerViewRowWithTitle:(NSString *)title inComponent:(NSInteger)co
...
+                // check if the month and day of the component is found in the column and set the components respectively.
+                if ([self.class.__kifDateFormatter.monthSymbols containsObject:component]) {
+                    dateComponents.month = [self.class.__kifDateFormatter.monthSymbols indexOfObject:component] + 1;
+                } else if ([self.class.__kifDateFormatter.shortMonthSymbols containsObject:component]) {
+                    dateComponents.month = [self.class.__kifDateFormatter.shortMonthSymbols indexOfObject:component] + 1;
+                } else if ([component integerValue] != 0) { // days can't be zero and if a string is not a number it will default to zero.
+                    dateComponents.day = [component integerValue];
+                }
+            }
+        }
```
justinseanmartin, line 788: Should there be an else with a throw here for unsupported mode?

Classes/KIFUIViewTestActor.m
```diff
@@ -465,6 +466,27 @@ - (void)selectPickerViewRowWithTitle:(NSString *)title inComponent:(NSInteger)co
     }
 }
 
+#pragma mark - Date Picker Actions
+
+- (void)selectDatePickerDate:(NSDate *)date
+{
+    @autoreleasepool {
+        KIFUIObject *found = [[self _usingExpectedClass:[UIDatePicker class]] _predicateSearchWithRequiresMatch:NO mustBeTappable:NO];
```
justinseanmartin, line 474: Can we add a predicate to ensure `datePickerMode` is the right type here and below, rather than asserting if we match on the wrong object?

justinseanmartin (approved):
This looks great, thanks!

Classes/KIFUITestActor.m
dostrander: If it's anything else it means it's a countdown timer which shouldn't do anything with date components

Classes/KIFUIViewTestActor.m
dostrander: Isn't that what we are doing with `[self _usingExpectedClass:[UIPickerView class]]`

PR #1192, opened 2020-11-09
Have the KIF visualizer not become first responder in the responder chain

dostrander:
This fixes an issue in the responder chain where textfields would sometimes not become the first responder if the window was not present


harleyjcooper (approved):
LGTM

PR #1195, opened 2020-11-24
Add visualizer to top window instead of a seperate window

dostrander:
This Fixes an issue that arises when switching windows and changing first responder due to visualizer.

### DIFF
--- a/Visualizer/KIFEventVisualizer.m
+++ b/Visualizer/KIFEventVisualizer.m
@@ -5,20 +5,6 @@
 #import "KIFEventVisualizer.h"
 #import "KIFTouchVisualizerViewCoordinator.h"
 
-static BOOL __isKIFVisualizerCreated = NO;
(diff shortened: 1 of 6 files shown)


Visualizer/KIFEventVisualizer.m
```diff
@@ -5,40 +5,20 @@
...
-
-- (UIWindowLevel)windowLevel
-{
-    return 100000099;
-}
-
-@end
-
-@interface KIFEventVisualizer()
+es@interface KIFEventVisualizer()
```
harleyjcooper, line 8: ?

dostrander: must have accidently typed something in there before commiting

harleyjcooper (approved):
👍

justinseanmartin (approved):

PR #1208, opened 2021-02-02
Wait for keyboard input tasks to finish before continuing typing

dostrander:
Making sure that the keyboard "flushes" on each keystroke to account for slower devices/simulators that may take longer to run or if a device is doing a lot of other processes and is slowing down the machine the keystroke will still get processed before "moving on" with the test, possibly causing errors within the test.

### DIFF
--- a/Sources/KIF/Classes/KIFTypist.m
+++ b/Sources/KIF/Classes/KIFTypist.m
@@ -13,10 +13,15 @@
 #import "CGGeometry-KIFAdditions.h"
 #import "UIAccessibilityElement-KIFAdditions.h"
 
+@interface UIKeyboardTaskQueue: NSObject
+- (void)waitUntilAllTasksAreFinished;
+@end
+
 @interface UIKeyboardImpl : NSObject
 + (UIKeyboardImpl *)sharedInstance;
 - (void)addInputString:(NSString *)string;
 - (void)deleteFromInput;
+@property(readonly, nonatomic) UIKeyboardTaskQueue *taskQueue;
 @property(getter=isInHardwareKeyboardMode) BOOL inHardwareKeyboardMode;
 @property(retain) UIResponder<UIKeyInput> * delegate;
 @end
@@ -89,6 +94,7 @@ static NSTimeInterval keystrokeDelay = 0.01f;
         [[UIKeyboardImpl sharedInstance] addInputString:characterString];
     }
     
+    [[[UIKeyboardImpl sharedInstance] taskQueue] waitUntilAllTasksAreFinished];
     CFRunLoopRunInMode(kCFRunLoopDefaultMode, keystrokeDelay, false);
     return YES;
 }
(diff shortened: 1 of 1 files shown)


justinseanmartin:
Do we want to do this on every character input or after the full string is entered? I haven't really seen issues with key presses coming out of order, so seems like maybe we only need to wait after all keys are entered. 

What is the performance overhead for this on typing a string? Can we remove the `keystrokeDelay` afterwards? Or would it make sense to switch the order of these calls?

dostrander:
> Do we want to do this on every character input or after the full string is entered? I haven't really seen issues with key presses coming out of order, so seems like maybe we only need to wait after all keys are entered.
> What is the performance overhead for this on typing a string? Can we remove or reduce the `keystrokeDelay`? Alternatively, does it make sense to switch the order of these calls?
I don't think we want to remove keystrokeDelay because IIRC it's settable.  
I think we want to keep it in between and BEFORE calling delay so if there are long running things that happen in the delegate (which IMO shouldn't but to each their own) it will still have the delay but AFTER the keyboard has "finished" typing, if that makes sense.

justinseanmartin:
> > Do we want to do this on every character input or after the full string is entered? I haven't really seen issues with key presses coming out of order, so seems like maybe we only need to wait after all keys are entered.
> > What is the performance overhead for this on typing a string? Can we remove or reduce the `keystrokeDelay`? Alternatively, does it make sense to switch the order of these calls?
> 
> I don't think we want to remove keystrokeDelay because IIRC it's settable.
> I think we want to keep it in between and BEFORE calling delay so if there are long running things that happen in the delegate (which IMO shouldn't but to each their own) it will still have the delay but AFTER the keyboard has "finished" typing, if that makes sense.

Response seems reasonable. My ask would be to check and see how much this slows test execution by overall, as this is a pretty hot path for tests, and could add up pretty substantially across the entire suite. Maybe we're able to reduce the `keystrokeDelay` some amount to offset the additional time?

dostrander:
> > > Do we want to do this on every character input or after the full string is entered? I haven't really seen issues with key presses coming out of order, so seems like maybe we only need to wait after all keys are entered.
> > > What is the performance overhead for this on typing a string? Can we remove or reduce the `keystrokeDelay`? Alternatively, does it make sense to switch the order of these calls?
> > 
> > 
> > I don't think we want to remove keystrokeDelay because IIRC it's settable.
> > I think we want to keep it in between and BEFORE calling delay so if there are long running things that happen in the delegate (which IMO shouldn't but to each their own) it will still have the delay but AFTER the keyboard has "finished" typing, if that makes sense.
> 
> Response seems reasonable. My ask would be to check and see how much this slows test execution by overall, as this is a pretty hot path for tests, and could add up pretty substantially across the entire suite. Maybe we're able to reduce the `keystrokeDelay` some amount to offset the additional time?

Just tested with a 2000 character string and there was a .1s difference in `[KIFTypist enterCharacter]`

We can reduce the default keystrokeDelay

justinseanmartin (approved):

PR #1209, opened 2021-02-02
Readd xcattachment screenshots

dostrander:
We only officially support Xcode 11+ so we can update re-add xcattachment screenshots everywhere


harleyjcooper (approved):

justinseanmartin (approved):
Should probably at least do a minor version bump for this with explicit release notes about dropping Xcode 10 support.

PR #1215, opened 2021-02-24
add IOKit in linker settings in Swift PM

dostrander:
Causes errors in project if you don't have IOKit linked


justinseanmartin:
Can we get the sample that wires up using SPM to be built on CI to validate this in the future?

justinseanmartin (approved):

Issue #1214, opened 2021-02-23
Selecting date picker value when picker timezone is not local does not give the expected time


danielbarela:
We have a UIDatePicker which has the timeZone set to ```TimeZone(secondsFromGMT: 0)``` We then test this by calling ```tester().selectDatePickerValue(["Nov 2", "7", "00", "AM"], with: .forwardFromCurrentValue);``` which ultimately should set the date to November 2 7:00:00 +0000, however it is being set to November 2 14:00:00 +0000 because my timezone is 7 hours behind GMT.

If I modify the KFUITestActor class in the ```- (void)selectDatePickerValue:(NSArray *)datePickerColumnValues fromPicker:(UIDatePicker *)datePicker withSearchOrder:(KIFPickerSearchOrder)searchOrder``` method to set the dateComponents timeZone to datePicker.timeZone, it works as expected.  Would you like me to issue a pull request for this?

dostrander:
@danielbarela if you could that would be great. If not i'll try to get to it in the next few weeks

PR #1216, opened 2021-02-24
Sets the timezone when selecting a date picker value

danielbarela:
The NSDateComponents object which is used to select the date picker values needs to reflect the timeZone set on the datePicker to properly select the values.

Fixes https://github.com/kif-framework/KIF/issues/1214


dostrander:
The fix looks right. When you get chance to fix up the builds and add a test i think we should be able to merge this. Thanks for doing this! :)

danielbarela:
I fixed the build issue, Xcode had modified the storyboard file since I use a newer version. My initial pull request had a test for the updated code. Let me know if I need to fix up anything else. Thanks

dostrander (approved):

dostrander:
Looks good, merging!

Issue #1212, opened 2021-02-22
KIF swift package manager warnings and errors


kikeenrique:
Hi 😀!,
I was eager to try SPM and I've just tested it with version 3.8.0 and I get a warning and some errors.

1. I created an empty project on Xcode, with testing artefacts.  
2. Added KIF as dependency on "Swift Packages"
3. Hit tests.

As a result, I get this warning:
```
Umbrella directory '/Users/kikeenrique/Library/Developer/Xcode/DerivedData/Iris-ipad-eqvhyhvbervzgndvqpelptfppwhj/SourcePackages/checkouts/KIF/Sources/KIF/include' not found
```
inside this autogenerated KIF.modulemap.
```
module KIF {
umbrella "/Users/kikeenrique/Library/Developer/Xcode/DerivedData/TestKIF-cyhhveqlzbzhwmflkpwnrpiuptwz/SourcePackages/checkouts/KIF/Sources/KIF/include"
export *
}
```

And on linking phase I see this errors:
```
Undefined symbols for architecture x86_64:
  "_IOHIDEventCreateDigitizerEvent", referenced from:
      _kif_IOHIDEventWithTouches in KIF.o
  "_IOHIDEventCreateDigitizerFingerEventWithQuality", referenced from:
      _kif_IOHIDEventWithTouches in KIF.o
  "_IOHIDEventSetIntegerValue", referenced from:
      _kif_IOHIDEventWithTouches in KIF.o
  "_IOHIDEventAppendEvent", referenced from:
      _kif_IOHIDEventWithTouches in KIF.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Regards

kenji21:
to fix, the undefined symbols, you may add `linkerSettings: [.linkedFramework("IOKit")]),` (like here: https://github.com/openium/KIF/blob/spm-support-freez/Package.swift#L23)

```
        .target(
            name: "KIF",
            dependencies: [],
            cSettings: [.headerSearchPath("ApplePrivateAPIs/")],
            linkerSettings: [.linkedFramework("IOKit")]),
```

kikeenrique:
@kenji21 you mean that fix must be applied in KIF, don't you?

About the umbrella module map, I haven't found where is it being generated, so no clue on how to fix it. Anyone?

kenji21:
[...]

PR #1219, opened 2021-03-25
SPM Integration Fixes

pbartolome:
This PR fixes the issues commented on https://github.com/kif-framework/KIF/issues/1212

- Public headers have been symlinked to a `include` folder, which is listed in the `publicHeaders` section of the `Package.swift`. Now clients can access classes that are inside the KIF module.
- No more warnings on the umbrella header
- Add `IOKit` as a requirement when using SPM
- Added a new sample project `SPMIntegration` which uses SPM to add `KIF` to a Swift test target


dostrander (commented):
Looks good. The only thing is can we move the SPMIntegration project into the root folder instead of the `Documentation/Examples` folder?

Thanks for doing this

pbartolome:
There was an existing failing tests `testSelectingCurrentDateTime`... interestingly it was failing at a specific time (if executed around 12pm). There was an issue in the test actor using the picker, incorrectly adding 12 when the time was 12pm to make it 24h format. I pushed a fix and a test so it doesn't add 12 when the time is 12pm.

```
10am + 0 = 10 OK
11am + 0 = 11 OK
12pm + 12 = 24 INCORRECT, should be 12pm + 0 = 12
1pm + 12 = 13 OK
2pm + 12 = 14 OK
```

dostrander (approved):

kenji21:
👍🏻 (remember that a "server" tag needs to be released to make this available via SPM)

pbartolome:
Thanks for merging it, any plans to release a new version with the changes?

dostrander:
Yeah will do this week

kikeenrique:
I just miss an updated readme reflecting the support/instructions for SPM.

dostrander:
Released a new version.

PR #1232, opened 2021-07-16
Fixed table view scroll behavior

till0xff:
Old implementation does not restore the initial offset in the table view after a view search


justinseanmartin (approved):
@dostrander - This might fix the issue that @jbeachwood mentioned the other day about scroll position.

dostrander (approved):

dostrander:
@till0xff thanks!

PR #1236, opened 2021-09-14
Audit timeouts and use executionBlockTimeout as default timeout

dostrander:
This audits the use of timeouts and uses the consistent `executionBlockTimeout` where magic numbers are specified. This makes it consistent that when a timeout is specified it is used throughout KIF.


dostrander:
Declining for now to come up with a better pattern for this

PR #1245, opened 2021-11-03
Check that we have the same view after scrolling

dostrander:
The actual view can be changed due to cell reuse after scrolling. This change ensures that once we scroll we have the same view as before we did so the action being taken is correct


justinseanmartin (approved):
Seems reasonable. Is there a test we could add that shows this being broken and then fixing it? Or is the lift to set up the specific scenario too prohibitive?

dostrander:
@justinseanmartin i can look into but it seems somewhat not consistent

justinseanmartin:
> @justinseanmartin i can look into but it seems somewhat not consistent

Was just a thought, not a blocker.

PR #1248, opened 2021-11-12
Add identifier to printing view hierarchy

dostrander:
Allows for better visibility when you are debugging the views. Folks can sometimes have a label and an identifier, however be looking up an identifier. This allows for better understanding of if that identifier is on screen.


justinseanmartin (approved):
Only downside I see is it loses some of the directional nudging to use label (visible to accessibility tools) over identifier (dev only). That said, maybe KIF should be less opinionated in that stance. It is unfortunate to erode some of that, but it is also the direction I've seen the framework taking.

harleyjcooper (approved):

PR #1249, opened 2021-11-12
Check full view when scrolling through Collectionview

dostrander:
This checks the full collectionview on each iteration of a cell, without scrolling, to ensure we don't miss an accessibility element that is on screen but is not in the cell.

**Why**
To look through a collection we iterate through the cells and look for the element in the cell. This is a good approximation for collectionviews as they can scroll horizontal or vertical so we can't assume to only scroll a single direction. However, our current implementation would only stop once the cell matched the accessibility attributes passed in. This would miss other elements on the screen like supplementary or decoration views that were on screen. This change instead does a non-scrolling lookup of the full view on each cell iteration to determine if the element is on screen.


harleyjcooper (approved):
LGTM as long as the increased delay won't potentially cause massive slowdowns?

PR #1251, opened 2021-12-08
Fix bug with table view elements that are matched but offscreen

dostrander:
This re-introduces https://github.com/kif-framework/KIF/pull/1070 where elements off screen in a scrollview need to be scrolled before they are tapped.

In iOS 15 some elements in a tableview are within the scrollview though not on screen, specifically table view footers (not to be confused with table view section footers). This causes an element to be matched and "found" but untappable.  Adding this will make sure that we scroll to the element before tapping it

### DIFF
--- a/Sources/KIF/Classes/KIFUITestActor.m
+++ b/Sources/KIF/Classes/KIFUITestActor.m
@@ -326,17 +326,7 @@ static BOOL KIFUITestActorAnimationsEnabled = YES;
 
         // If the element isn't immediately tappable, try checking if it is contained within scroll views that can be scrolled to make it tappable.
         if (isnan(tappablePointInElement.x)) {
-            UIView *container = view;
-
-            do {
-                if ([container isKindOfClass:UIScrollView.class]) {
-                    UIScrollView *containerScrollView = (UIScrollView *)container;
-                    CGRect rect = [view convertRect:view.frame toView:containerScrollView];
-                    [containerScrollView scrollRectToVisible:rect animated:NO];
-                }
-
-                container = container.superview;
-            } while (container != nil);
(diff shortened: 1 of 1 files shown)


harleyjcooper (approved):

babbage:
This commit broke a substantial number of my tests, which used KIF to navigate the UI but also included snapshot testing. When `tester().tapView(withAccessibilityLabel: "My accessibility label")` is called, following this commit any scroll view is scrolled to it's bottom bounds, even where the item that is being waited for was already visible when the method was called. This results in scroll view displacement which in my case caused snapshot tests to fail as the new screenshots did not match with the expected baselines in many cases.

I am removing these `waitForView...` calls as they no longer seem to be necessary for my tests to run reliably anyway, which was the reason they were originally added. I do wonder however about whether this commit has overshot its intention given what I have observed. It would seem to me that KIF should first attempt to ascertain if the view being requested is visible _without scrolling_ and should only do this new auto-scrolling (if at all) when there is a problem with the first approach.

It may be I have not understood the original intentions of this patch... I did not attempt to take in the full picture, just enough to confirm this definitely appears to be the cause of many of my snapshot tests expectedly failing.

dostrander:
@babbage What you are describing is a side effect of [this PR](https://github.com/kif-framework/KIF/pull/1246). Which we've actually reverted [here](https://github.com/kif-framework/KIF/pull/1252) and are planning to cut a release once we are able to merge 1 more PR for iOS 15. Sorry for the mishap

babbage:
@dostrander yes, that'd be it! My issue was indeed seen in UICollectionViews. Thanks for the update. :)

PR #1252, opened 2021-12-08
Revert "Merge pull request #1246 from kif-framework/derko/mimic-table…

dostrander:
…view-element-matching"

This reverts commit aa8d173f698c35c640b558f855c44d5cd9395f59, reversing
changes made to 252b34ecb40987e7476a2aaaead7a73099de4d9b.

**Why**
Collectionview's render different than tableview's. A cell isn't able to be rendered unless the main runloop runs and it is supposed to be on screen. This forces us to run the runloop and actually scroll the view within the forloop of looking for the element in the scrollview. This causes un-intended side effects because users of this might change the UI based on if the collectionview scrolls which can cause bugs and flakiness depending on the ordering of searching the view hierarchy. For example if you have a label and a collectionview in a view and KIF searches the view heirarchy, finding the collectionview first it will scroll and MAY change what the label says, which isn't exactly what we want.

Overall i think we can re-introduce this if we make 2 somewhat large changes to KIF. 
First change the priority of what get's searched first. Namely we deprioritize searching scrollview's until the end. This may require us to change the implementation of searching to not be recursive.
Second we introduce an API to enforce that we do not scroll any scrollview's on screen.

### DIFF
--- a/Sources/KIF/Additions/UIView-KIFAdditions.m
+++ b/Sources/KIF/Additions/UIView-KIFAdditions.m
@@ -303,50 +303,48 @@ NS_INLINE BOOL StringsMatchExceptLineBreaks(NSString *expected, NSString *actual
         } else if ([self isKindOfClass:[UICollectionView class]]) {
             UICollectionView *collectionView = (UICollectionView *)self;
             
-            NSMutableArray *indexPathsForVisibleItems = [[NSMutableArray alloc] init];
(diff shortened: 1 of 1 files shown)


justinseanmartin (approved):
I don't think the second proposal is a ton of work, but reasonable to back this out in the meantime. We might even consider making scrolling of scroll views an opt-in behavior (breaking change, would require batching with major version bump).

harleyjcooper (approved):

PR #1253, opened 2021-12-08
Directly Specify the OS versions and move Xcode 10 to 10.15

dostrander:
In github actions it looks like macos-latest has changed to 11. This specifies only Xcode 10 should run on 10.15 and everything else should run on 11


justinseanmartin:
Bummer that we can't test Monterey yet: https://github.com/github/roadmap/issues/240

PR #1268, opened 2022-09-20
Add Xcode 14 and iOS 16 running in KIF

dostrander:


justinseanmartin:
Finally done in #1300

PR #1276, opened 2023-02-19
Fix ios 16

BrunoMazzo:
Fix for iOS 16. 

- Adding the pickerView into the inputView of a textfield is handle different than the other iOS versions. On iOS 16, it is part of the property `_fallbackView` from `_UIRemoteKeyboardPlaceholderView`, but it is not part of the UI hierarchy (I don't know why). So in the parts that checked for elements, I added another validation for the `_UIRemoteKeyboardPlaceholderView`
- UIDatePicker is not a subclass of UIPickerView. And it has doesn't has an UIPickerView as subview. The hierarchy is UIDatePicker -> _UIDatePickerView ->  ... -> UIPickerColumnView ... But it has an _UIPickerView property that works if used.
- `[[UIDevice currentDevice] setValue:@(orientation) forKey:@"orientation"]` doesn't work anymore. We need to use `requestGeometryUpdateWithPreferences` from UIWindowScene.

I'm not 100% familiar with the code, so I may be missing some context, so please let me know if there is anything I should fix.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,6 +15,43 @@ jobs:
     - name: Validate podspec
       run: pod lib lint
 
+  build_xcode14:
+    runs-on: macos-12
+    strategy:
+      matrix:
+        run-config:
+          - { xcode_version: '14.1', simulator: 'name=iPad Pro (12.9-inch) (5th generation),OS=16.1' }
+    steps: &build_steps
+    - name: Checkout Project
+      uses: actions/checkout@v1
+
+    - name: Brew Update
+      run:  brew update
+
+    - name: Install Bundler
+      run: gem install bundler
+
+    - name: Install Core Utils
+      run: if [ -z "$(brew ls --versions coreutils)" ] ; then brew install coreutils ; fi
+
+    - name: Install XCPretty
+      run: gem install xcpretty --no-document --quiet
+
+    - name: Show Xcode versions
+      run: ls -al /Applications/Xcode*
+
+    - name: Select Xcode
+      run: sudo xcode-select -s /Applications/Xcode_${{ matrix.run-config['xcode_version'] }}.app
+
+    - name: Current Xcode Selected
+      run: xcode-select -p
+
+    - name: List Simulators
+      run:  xcrun simctl list
+
+    - name: Build & Test
+      run: ./scripts/ci.sh "${{ matrix.run-config['simulator'] }}"
+
   build:
     runs-on: macos-11
     strategy:
@@ -23,38 +60,7 @@ jobs:
           - { xcode_version: '11.7', simulator: 'name=iPhone SE (2nd generation),OS=13.7' }
           - { xcode_version: '12.5.1', simulator: 'name=iPhone SE (2nd generation),OS=14.5' }
           - { xcode_version: '13.0', simulator: 'name=iPad Pro (12.9-inch) (5th generation),OS=15.0' }
-
-
-    steps:
-    - name: Checkout Project
-      uses: actions/checkout@v1
-
-    - name: Brew Update
-      run:  brew update
-
-    - name: Install Bundler
-      run: gem install bundler
-
-    - name: Install Core Utils
-      run: if [ -z "$(brew ls --versions coreutils)" ] ; then brew install coreutils ; fi
-
-    - name: Install XCPretty
-      run: gem install xcpretty --no-document --quiet
-
-    - name: Show Xcode versions
-      run: ls -al /Applications/Xcode*
-
-    - name: Select Xcode
-      run: sudo xcode-select -s /Applications/Xcode_${{ matrix.run-config['xcode_version'] }}.app
-
-    - name: Current Xcode Selected
-      run: xcode-select -p
-
-    - name: List Simulators
-      run:  xcrun simctl list
-
-    - name: Build & Test
-      run: ./scripts/ci.sh "${{ matrix.run-config['simulator'] }}"
+    steps: *build_steps
 
   build_xcode10:
     runs-on: macos-10.15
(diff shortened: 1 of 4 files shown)


Sources/KIF/Additions/NSError-KIFAdditions.h
```diff
@@ -13,4 +13,6 @@
 + (instancetype)KIFErrorWithUnderlyingError:(NSError *)underlyingError format:(NSString *)format, ... NS_FORMAT_FUNCTION(2,3);
 + (instancetype)KIFErrorWithFormat:(NSString *)format, ... NS_FORMAT_FUNCTION(1,2);
 
++ (instancetype)KIFErrorFromException:(NSException *)exception;
```
justinseanmartin, line 16: I think this got pulled over from the other PR. Worth reverting here, as it seems unused in this context.

Sources/KIF/Additions/UIView-KIFAdditions.m
```diff
@@ -374,12 +382,25 @@ - (NSArray *)subviewsWithClassNamePrefix:(NSString *)prefix;
         if ([NSStringFromClass([view class]) hasPrefix:prefix]) {
             [result addObject:view];
         }
+        
+        if([NSStringFromClass([view class]) isEqualToString:@"_UIRemoteKeyboardPlaceholderView"]) {
```
justinseanmartin, line 386: Rather than duplicating this logic, it'd be cleaner to add a `subviewsWithPlaceholderFallbackView` method that can be used in place of `self.subviews` in the 4 places we're iterating over `self.subviews` in this file.

Sources/KIF/Classes/KIFSystemTestActor.m
```diff
@@ -58,7 +58,50 @@ - (void)simulateMemoryWarning
...
+                    orientationMask = UIInterfaceOrientationMaskAll;
+                    break;
+                case UIDeviceOrientationFaceDown:
+                    orientationMask = UIInterfaceOrientationMaskAll;
+                    break;
+            }
+            
+            UIWindowSceneGeometryPreferencesIOS* preferences = [[UIWindowSceneGeometryPreferencesIOS alloc]initWithInterfaceOrientations:orientationMask];
+            [windowScene requestGeometryUpdateWithPreferences:preferences errorHandler:^(NSError * _Nonnull error) {
+                NSLog(@"error: %@", error);
```
justinseanmartin, line 99: Does this happen in a expected case? If not, makes sense to emit an error and fail the test.

Sources/KIF/Classes/KIFUIViewTestActor.h
```diff
@@ -137,33 +137,44 @@ extern NSString *const inputFieldTestString;
  @abstract Tap a view matching the tester's search predicate.
  @discussion The tester will evaluate the accessibility hierarchy against it's search predicate and perform a tap on the first match.
  */
-- (void)tap;
+- (void)tap CF_SWIFT_UNAVAILABLE_FROM_ASYNC("use async version");
```
justinseanmartin, line 140: This should be reverted and decoupled from then changes in this PR I believe.

justinseanmartin (commented):
Thank you so much for your contribution! 2 minor improvement requests and we should remove the commits that are also included in #1275. This is great!

Does this get the full KIF integration test suite passing on iOS 16 devices? If so, it would be great to get CI validating that. If not, maybe we should start working towards that and explicitly bailing out for the specific tests that we know are broken on iOS 16.

BrunoMazzo:
Thanks for the feedback. I will work on it and push the changes probably in the weekend. 

This fix the full suite of tests. I will remove the changes for the async stuff that I forgot (I starting work in the same branch and forgot to clean up), add the CI validation, and do the other changes.

justinseanmartin:
That's great to hear! It is probably worth bumping the CI config to validate that we don't regress it. That's definitely not a blocker for landing these changes though.

.github/workflows/build.yml
```diff
@@ -15,17 +15,13 @@ jobs:
...
       matrix:
         run-config:
-          - { xcode_version: '11.7', simulator: 'name=iPhone SE (2nd generation),OS=13.7' }
-          - { xcode_version: '12.5.1', simulator: 'name=iPhone SE (2nd generation),OS=14.5' }
-          - { xcode_version: '13.0', simulator: 'name=iPad Pro (12.9-inch) (5th generation),OS=15.0' }
-
-
-    steps:
+          - { xcode_version: '14.1', simulator: 'name=iPad Pro (12.9-inch) (5th generation),OS=16.1' }
+    steps: &build_steps
```
justinseanmartin, line 24: Bummer... I liked the cleanup with the anchor, but GitHub doesn't like it:

```
The workflow is not valid. .github/workflows/build.yml: Anchors are not currently supported. Remove the anchor 'build_steps'
```

justinseanmartin (commented):
Tour contributions are very much appreciated! If we can get CI green, then we should be good to go.

.github/workflows/build.yml
BrunoMazzo: My bad. I will fix it and push it right now!

dostrander:
@BrunoMazzo thanks 🎉 for finding this! So this allows you to use pickers and keyboads in iOS 16??

I was under the impression this was out of process from my debugging in this so was expecting a larger reverse engineering effort to be able to call what XCUITest calls. But if that's not that case that's amazing. Thank you SOOO much for digging into this!

dostrander:
If you need some help getting it to run on iOS 16 feel free to use anything from here: https://github.com/kif-framework/KIF/pull/1268

I added this a while ago but since the tests weren't passing it didn't make sense to merge

justinseanmartin:
I'm rerunning the failed jobs to see if the results are consistent or not. I think at this point, supporting only Xcode 13.4.1 (Monterey) & 14.2 (Ventura) seems reasonable to me. Maybe we could do one iPad + iPhone device on each platform? Not changing the device type will help eliminate some variable that might lead other test failures related to screen dimensions and whatnot. WDYT? Just trying to limit how many things we need to chase down here to get this landed.

justinseanmartin:
It looks like we might need some availability checking for the behavior to be backwards compatible with the iOS 15 (Xcode 13) SDK:

```
❌  /Users/runner/work/KIF/KIF/Sources/KIF/Classes/KIFSystemTestActor.m:98:26: no visible @interface for 'UIWindowScene' declares the selector 'requestGeometryUpdateWithPreferences:errorHandler:'
```

BrunoMazzo:
I added a check to compile on other versions of Xcode and changed the CI to run on iPhone SE and iPad Pro, Xcode 13 and 14.

.github/workflows/build.yml
```diff
@@ -15,16 +15,13 @@ jobs:
...
     strategy:
       matrix:
         run-config:
-          - { xcode_version: '11.7', simulator: 'name=iPhone SE (2nd generation),OS=13.7' }
-          - { xcode_version: '12.5.1', simulator: 'name=iPhone SE (2nd generation),OS=14.5' }
-          - { xcode_version: '13.0', simulator: 'name=iPad Pro (12.9-inch) (5th generation),OS=15.0' }
-
-
+          - { xcode_version: '14.1', simulator: 'name=iPad Pro (12.9-inch) (5th generation),OS=16.1' }
+          - { xcode_version: '14.1', simulator: 'name=iPhone SE (3nd generation),OS=16.1' }
```
justinseanmartin, line 24: Meant for this to be 2nd instead of the 3nd?

justinseanmartin (approved):
Once CI is green, we should be good to go with this.

PR #1278, opened 2023-03-30
Support detecting a view is offscreen

steve-sun-me:
- Fix an issue when a view is off a screen completely but not hidden, KIF thinks it's still there
- Added the `alpha` condition check. When a view's alpha is 0, we treat is as absent


Sources/KIF/Additions/UIView-KIFAdditions.m
```diff
@@ -1062,4 +1099,20 @@ -(UIView*)tryGetiOS16KeyboardFallbackViewFromParentView:(UIView*) parentView {
     return nil;
 }
 
+- (nullable UIView *)parentPlainScrollView
+{
+    UIView *currentSuperview = self.superview;
+    while (currentSuperview != nil) {
+        if ([currentSuperview isKindOfClass:[UIScrollView class]] &&
+            ![currentSuperview isKindOfClass:[UICollectionView class]] &&
```
dostrander, line 1107: Probably want a comment in explaining why we don't want uicollectionviews and uitableviews

Sources/KIF/Additions/UIView-KIFAdditions.m
```diff
@@ -1062,4 +1099,22 @@ -(UIView*)tryGetiOS16KeyboardFallbackViewFromParentView:(UIView*) parentView {
     return nil;
 }
 
+- (nullable UIView *)parentPlainScrollView
+{
+    // We don't want collection view and table view because we handle them separately.
+    // This function is only getting a plain scroll view
+    UIView *currentSuperview = self.superview;
```
justinseanmartin, line 1106: What about the scenario where the element being tapped on is a descendant rather than a direct subview of a scroll view? In more complex controls, I'd think this would happen relatively often.

Sources/KIF/Classes/KIFUIViewTestActor.m
```diff
@@ -197,7 +197,7 @@ - (void)waitForAbsenceOfView;
         KIFTestWaitCondition(found.view, error, @"Cannot find view containing accessibility element \"%@\"", found.element);
         
         // Hidden views count as absent
-        KIFTestWaitCondition([found.view isHidden] || [found.view superview] == nil, error, @"Accessibility element \"%@\" is visible and not hidden.", found);
+        KIFTestWaitCondition([found.view isHidden] || [found.view superview] == nil || ![found.view isPossibleVisibleInWindow], error, @"Accessibility element \"%@\" is visible and not hidden.", found);
```
justinseanmartin, line 200: It feels odd doing this in `waitForAbsenceOfView`. Would this same view be usable for `waitForView` or `tap` actions? Potentially not the latter, due to maybe failing the `mustBeTappable:YES` param.

Sources/KIF/Additions/UIView-KIFAdditions.m
steve-sun-me: Trying to picture this in my head, can you provide an example?

dostrander: IF it's a `UIScrollview` that contains a `UIView` that contains a `UILabel` for instance`

So something visually like:

```
> UIScrollView
--> UIView
---->UILabel ** This one
--> UITextView
```

justinseanmartin: Yep, something like that is what I was thinking

Sources/KIF/Classes/KIFUIViewTestActor.m
justinseanmartin: FYI - not outdated, just marked that way due to the method name change.

steve-sun-me: > Would this same view be usable for waitForView or tap actions?

From what I followed the code path it doesn't seem like it's being used in `waitForView` or `tap`

Sources/KIF/Additions/UIView-KIFAdditions.m
steve-sun-me: @justinseanmartin I think I am already doing that for the `while loop` no?

justinseanmartin: Yep, you're right, sorry I misunderstood. Would recommend using the name `ancestor` here to make it a bit more clear.

KIF Tests/OffscreenTests.m
```diff
@@ -0,0 +1,36 @@
...
+//  OffscreenTests.m
+//  KIF Tests
+//
+//  Created by Steve Sun on 2023-03-28.
+//
+
+#import <KIF/KIF.h>
+#import "KIFTestStepValidation.h"
+
+@interface OffscreenTests : KIFTestCase
```
justinseanmartin, line 11: We should either add an equivalent version of this against the KIFUIViewTestActor, or use that version, as that is what we'd like to be the first class API.

KIF Tests/OffscreenTests.m
```diff
@@ -0,0 +1,36 @@
...
+{
+    [tester tapViewWithAccessibilityLabel:@"Offscreen Views"];
+}
+
+- (void)afterEach
+{
+    [tester tapViewWithAccessibilityLabel:@"Test Suite" traits:UIAccessibilityTraitButton];
+}
+
+- (void)testViewOffscreen
```
justinseanmartin, line 26: We should probably move the views back into the scrollable region and unhide them, and make sure that they pass the `waitForView` checks. While they're offscreen, they we should probably be able to ensure that they fail the `waitForView` checks.

justinseanmartin (approved):

steve-sun-me:
@justinseanmartin - answering the [thread](https://github.com/kif-framework/KIF/pull/1278#discussion_r1153760124) here for viz. 

Originally `waitForAbsent`, `waitForView`, `tap` all have the bug where it returns true even the matching view is outside of visible area. This PR is addressing the issue for `waitForAbsent` only and leave `waitForView` and `tap` the same as is. I will come back to tackle them later

PR #1279, opened 2023-04-02
Set animation speed for all windows

steve-sun-me:
We want to apply the same animation speed to all windows

### DIFF
--- a/Sources/KIF/Additions/UIWindow+KIFSwizzle.m
+++ b/Sources/KIF/Additions/UIWindow+KIFSwizzle.m
@@ -34,7 +34,7 @@
 - (instancetype)swizzle_init
 {
     UIWindow *window = [self swizzle_init];
-    window.layer.speed = 1000;
+    window.layer.speed = [UIApplication sharedApplication].animationSpeed;
 
     return window;
 }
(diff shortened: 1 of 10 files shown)


dostrander (approved):

Sources/KIF/Additions/UIWindow+KIFSwizzle.m
```diff
@@ -0,0 +1,36 @@
...
+        SEL swizzledSelector = @selector(swizzle_becomeKeyWindow);
+
+        Method originalMethod = class_getInstanceMethod(class, originalSelector);
+        Method swizzledMethod = class_getInstanceMethod(class, swizzledSelector);
+
+        method_exchangeImplementations(originalMethod, swizzledMethod);
+    });
+}
+
+- (void)swizzle_becomeKeyWindow
```
justinseanmartin, line 30: Is there a scenario where we'd want this set in init as well, if a new window was created, but not made the keyWindow?

steve-sun-me: @justinseanmartin perhaps we only need to set this in init? Since all available UIWindows animation speed will get set due to [this](https://github.com/kif-framework/KIF/pull/1279/files#diff-02e0ae04f5d0c0f4f624ae5086b62b1812ade7722db159814bb1a1b4bae47b21R150)

steve-sun-me: Added both `init` and `initWithWindowScene`

PR #1283, opened 2023-05-01
Bump iOS minimum iOS deployment target to 11.0

danielob:
I've noticed that when opening the project with Xcode 14.3, we get a warning that the target is not in the range of supported versions. Additionally, when trying to build the project using Carthage, we get a build error due to the deployment target being set to 8.0.

To address this issue, I've made changes to update the deployment target version to 11.0, which is within the range of supported deployment target versions for both Xcode and Carthage. This will ensure that the project can be built and run smoothly with the latest versions of Xcode and Carthage.

I've tested the changes locally and confirmed that they resolve the warning and error messages. Please review my changes and let me know if you have any feedback or concerns.


justinseanmartin (approved):
Thank you for the contribution! The README already states that KIF officially supports iOS 11+, so this just makes it more explicit.

If you don't mind, it would be great to bump SPM and CocoaPods package definitions as well to match. If you don't want to scope creep, I can push that after your changes land.

danielob:
Looks like the test host app is using the deprecated `UIActionSheet`. It is recommended to use `UIAlertController` instead. I can try to make these changes and update the tests. Also, I noticed the KIF tests target is failing to compile because is using some old `openURL` API. I will try to also fix this.

adamsousa:
Any updates on merging this fix?

Issue #1281, opened 2023-04-04
Screen edge pan gesture is not recognized


Fitzmaz:
### Description
I am encountering difficulty in activating the built-in `interactivePopGestureRecognizer`. I have attempted to simulate a screen edge pan gesture using the `dragFromPoint:displacement:steps:` method, but the gesture is not being recognized.

### Code Snippet
```objc
CGFloat majorSwipeDisplacement = UIScreen.mainScreen.bounds.size.width * 0.5;
CGFloat minorSwipeDisplacement = 5;
KIFDisplacement displacement = CGPointMake(majorSwipeDisplacement, minorSwipeDisplacement);

[view dragFromPoint:CGPointMake(0.5, 150) displacement:displacement steps:20];
```

### Environment
- iOS version: 14.7, 16.1
- Xcode version: 14.1

Fitzmaz:
As mentioned in Pull Request #776, `interactivePopGestureRecognizer` was able to be triggered successfully. Looks like something might have changed in the iOS SDK.

dostrander:
@Fitzmaz I'd possibly try more steps and possibly start at 10-20 for x and middle of the screen for y. The maybe jump the steps to 100?

I don't have time currently to investigate this right now but if you can't repro or fix I'll try to get to it ASAP!

\cc @steve-sun-me

Fitzmaz:
@dostrander Thanks for the advices. I've experimented with various parameters, but none of them seem to work. So I don't think it's a problem with the parameters themselves.

Then, I did some digging and compared the `UITouch` object of a normal pan gesture with that of a screen edge pan gesture. It turns out that the touch object of the screen edge pan gesture has a property called `_edgeType` that is set, while the touch object of the normal pan gesture does not have this property set.

Regarding the KIF framework, touch objects are generated within the `dragPointsAlongPaths:` method. I discovered that if I manually set the `_edgeType` property of the touch object to a value of `4`, the screen edge pan gesture is recognized successfully. Maybe we should add a new method to the tester that can simulate a screen edge pan gesture and automatically set the touch object's property to the right value?

dostrander:
@Fitzmaz that seems like a decent solution to me.

One thing i find a little strange is why the edge pan gesture doesn't pick up the drag as it would normally? I'm curious how it would normally set that edge type under the hood

Fitzmaz:
[...]

PR #1286, opened 2023-06-29
Add support for swiping from the specified edge of the screen

Fitzmaz:
As discussed in issue #1281, I tried to address the problem by adding a new method to the tester. This method simulates a screen edge pan gesture by manually setting the `_edgeType` property. I wonder if there's any better solution, any feedback is greatly appreciated.

### DIFF
--- a/KIF Tests/GestureTests.m
+++ b/KIF Tests/GestureTests.m
@@ -193,6 +193,18 @@
     [tester waitForViewWithAccessibilityLabel:@"Down"];
 }
 
+- (void)testSwipeFromScreenEdgeLeft
+{
+    [tester swipeFromEdge:UIRectEdgeLeft atY:300];
+    [tester waitForViewWithAccessibilityLabel:@"LeftEdge"];
+}
+
+- (void)testSwipeFromScreenEdgeRight
+{
+    [tester swipeFromEdge:UIRectEdgeRight atY:300];
+    [tester waitForViewWithAccessibilityLabel:@"RightEdge"];
+}
+
 - (void)testScrolling
 {
     // Needs to be offset from the edge to prevent the navigation controller's interactivePopGestureRecognizer from triggering
--- a/KIF Tests/GestureTests.m
+++ b/KIF Tests/GestureTests.m
@@ -203,7 +203,7 @@
     [view dragFromPoint:point displacement:displacement steps:20];
     [tester waitForAbsenceOfViewWithAccessibilityLabel:@"LeftEdge"];
     
-    [tester swipeFromEdge:UIRectEdgeLeft atY:200];
+    [tester swipeFromEdge:UIRectEdgeLeft];
     [tester waitForViewWithAccessibilityLabel:@"LeftEdge"];
 }
 
@@ -217,7 +217,7 @@
     [view dragFromPoint:point displacement:displacement steps:20];
     [tester waitForAbsenceOfViewWithAccessibilityLabel:@"RightEdge"];
     
-    [tester swipeFromEdge:UIRectEdgeRight atY:200];
+    [tester swipeFromEdge:UIRectEdgeRight];
     [tester waitForViewWithAccessibilityLabel:@"RightEdge"];
 }
 
--- a/KIF Tests/GestureTests.m
+++ b/KIF Tests/GestureTests.m
@@ -193,15 +193,31 @@
     [tester waitForViewWithAccessibilityLabel:@"Down"];
 }
 
-- (void)testSwipeFromScreenEdgeLeft
+- (void)testSwipingFromScreenEdgeLeft
 {
-    [tester swipeFromEdge:UIRectEdgeLeft atY:300];
+    UIView *view = [tester waitForViewWithAccessibilityIdentifier:@"gestures.swipeMe"];
+    CGSize windowSize = view.window.bounds.size;
+    CGPoint point = CGPointMake(0.5, 200);
+    point = [view convertPoint:point fromView:view.window];
+    KIFDisplacement displacement = CGPointMake(windowSize.width * 0.5, 5);
+    [view dragFromPoint:point displacement:displacement steps:20];
+    [tester waitForAbsenceOfViewWithAccessibilityLabel:@"LeftEdge"];
+    
+    [tester swipeFromEdge:UIRectEdgeLeft atY:200];
     [tester waitForViewWithAccessibilityLabel:@"LeftEdge"];
 }
 
-- (void)testSwipeFromScreenEdgeRight
+- (void)testSwipingFromScreenEdgeRight
 {
-    [tester swipeFromEdge:UIRectEdgeRight atY:300];
+    UIView *view = [tester waitForViewWithAccessibilityIdentifier:@"gestures.swipeMe"];
+    CGSize windowSize = view.window.bounds.size;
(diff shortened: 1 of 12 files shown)


Sources/KIF/Additions/UIView-KIFAdditions.m
```diff
@@ -692,6 +701,7 @@ - (void)dragPointsAlongPaths:(NSArray<NSArray<NSValue *> *> *)arrayOfPaths {
                 point = [self convertPoint:point fromView:self.window];
                 UITouch *touch = [[UITouch alloc] initAtPoint:point inView:self];
                 [touch setPhaseAndUpdateTimestamp:UITouchPhaseBegan];
+                [touch setIsFromEdge:isFromEdge];
```
justinseanmartin, line 704: I wonder if we should alternatively come up with a heuristic here for what constitutes an edge swipe and do it implicitly. Being explicit seems reasonable as well though.

Sources/KIF/Additions/UITouch-KIFAdditions.m
```diff
@@ -117,6 +118,12 @@ - (void)setPhaseAndUpdateTimestamp:(UITouchPhase)phase
     [self setPhase:phase];
 }
 
+- (void)setIsFromEdge:(BOOL)isFromEdge
+{
+    NSInteger edgeType = isFromEdge ? 4 : 0;
+    [self _setEdgeType:edgeType];
+}
```
justinseanmartin, line 125: What would happen if we directly called `_UITouchSetBasicTouchPropertiesFromEvent`? Does Apple already have the appropriate heuristics to set this properly?

Sources/KIF/Additions/UIView-KIFAdditions.m
```diff
@@ -630,10 +630,15 @@ - (void)dragFromPoint:(CGPoint)startPoint toPoint:(CGPoint)endPoint steps:(NSUIn
 }
 
 - (void)dragFromPoint:(CGPoint)startPoint displacement:(KIFDisplacement)displacement steps:(NSUInteger)stepCount;
+{
+    [self dragFromPoint:startPoint displacement:displacement steps:stepCount isFromEdge:NO];
```
justinseanmartin, line 634: This is kind of a breaking change, in that it would require code changes in existing tests if they had working edge gestures. As you'd called out in the issue, this has been broken since iOS 11, so seems reasonable IMO.

KIF Tests/GestureTests.m
```diff
@@ -193,6 +193,18 @@ - (void)testSwipingDownWithIdentifier
     [tester waitForViewWithAccessibilityLabel:@"Down"];
 }
 
+- (void)testSwipeFromScreenEdgeLeft
```
justinseanmartin, line 196: Worth having a similar test for verifying that dragFromPoint doesn't trigger? Or is it reasonable to treat that as undefined behavior and be untested?

justinseanmartin (approved):
Couple of discussion points on the approach, but what you have here also seems reasonable. We appreciate the contribution and adding test coverage of the scenario.

Sources/KIF/Additions/UIView-KIFAdditions.m
dostrander: I think with this we can make an API specifically for swiping at an edge, similarly to like tapping on a button or status bar instead of using x,y coordinatotes. As well as implicitly doing it from the edge if we can detect it being around where it should be. I think that'd be in line with other APIs of KIf

Sources/KIF/Additions/UITouch-KIFAdditions.m
Fitzmaz: Apple derives the value of the property from the HIDEvent. We would need to synthesize a different HIDEvent for edge pan before calling `_UITouchSetBasicTouchPropertiesFromEvent`. This introduces additional complexity and I lack the necessary knowledge to implement it.

Sources/KIF/Additions/UIView-KIFAdditions.m
Fitzmaz: I thought about doing it implicitly like
```
if (startPoint.x < 1 || startPoint.x > screen.size.width - 1) {
    [touch setIsFromEdge:isFromEdge];
}
```
But this could be a breaking change for users who utilize the `dragFromPoint` method without intending to trigger a screen edge pan gesture.

Sources/KIF/Additions/UITouch-KIFAdditions.m
Fitzmaz: I wonder if we could replace these magic numbers with Apple enums.

Sources/KIF/Additions/UIView-KIFAdditions.m
Fitzmaz: I have tried to implement an alternative way based on the suggestions, which one do you prefer?

Sources/KIF/Additions/UIView-KIFAdditions.m
```diff
@@ -631,14 +631,30 @@ - (void)dragFromPoint:(CGPoint)startPoint toPoint:(CGPoint)endPoint steps:(NSUIn
...
+- (void)dragFromEdge:(UIRectEdge)startEdge toEdge:(UIRectEdge)endEdge
 {
-    CGPoint endPoint = CGPointMake(startPoint.x + displacement.x, startPoint.y + displacement.y);
-    NSArray<NSValue *> *path = [self pointsFromStartPoint:startPoint toPoint:endPoint steps:stepCount];
-    [self dragPointsAlongPaths:@[path] isFromEdge:isFromEdge];
+    CGFloat width = self.bounds.size.width;
+    CGFloat height = self.bounds.size.height;
+    CGFloat edgeInset = 0.5;
+    NSDictionary *edgeToPoint = @{
+        @(UIRectEdgeTop): @(CGPointMake(width / 2, 0)),
```
justinseanmartin, line 645: Should we also use `edgeInset` here for the top and `height - edgeInset` for the bottom?

justinseanmartin (commented):
Sorry for taking so long to get back here, and thanks for making the revisions! One other thing that we should do before landing this is add an equivalent API to `KIFUIViewTestActor`. That is the more flexible replacement of the `KIFUITestActor` API, though there is still a lot of documentation indicating to use the older API.

Sources/KIF/Additions/UIView-KIFAdditions.m
Fitzmaz: I'v made some revisions as suggested, please review again.

justinseanmartin (approved):

justinseanmartin:
Thanks again for the iterations here and the contribution! I'll get this merged once CI reports green.

justinseanmartin:
The tests appear to be extremely flaky right now in `AccessibilityIdentifierTests.testClearingAndEnteringTextIntoViewWithAccessibilityLabel`. I'm rerunning them and it's making progress on ensuring they're all green before merging.

dostrander:
> The tests appear to be extremely flaky right now in `AccessibilityIdentifierTests.testClearingAndEnteringTextIntoViewWithAccessibilityLabel`. I'm rerunning them and it's making progress on ensuring they're all green before merging.

Merged as they all went green

PR #1287, opened 2023-07-12
Fix occasional CI failures on the iPhone simulator

Fitzmaz:
In iOS, the edit menu is designed to display two commands and a "Forward" or "Back" page button when there are more than two commands available.

If the pasteboard already contains a string, the edit menu will show "Paste" and "Select" commands. The "Select All" command can be accessed on the next page.

Executing a case that taps the "Cut" command will result in the failure of all subsequent cases that require tapping the "Select All" command.

One solution to this is to clear the pasteboard before the "Select All" cases.

### DIFF
--- a/KIF Tests/AccessibilityIdentifierTests.m
+++ b/KIF Tests/AccessibilityIdentifierTests.m
@@ -17,6 +17,7 @@
 
 - (void)beforeEach
 {
+    UIPasteboard.generalPasteboard.string = nil;
     [tester tapViewWithAccessibilityLabel:@"Tapping"];
 }
 
--- a/KIF Tests/AccessibilityIdentifierTests_ViewTestActor.m
+++ b/KIF Tests/AccessibilityIdentifierTests_ViewTestActor.m
@@ -17,6 +17,7 @@
 
 - (void)beforeEach
 {
+    UIPasteboard.generalPasteboard.string = nil;
     [[viewTester usingLabel:@"Tapping"] tap];
 }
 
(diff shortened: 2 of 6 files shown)


dostrander (approved):

Fitzmaz:
Another failure in `AccessibilityIdentifierTests.testClearingAndEnteringTextIntoViewWithAccessibilityLabel`:

```
AccessibilityIdentifierTests
  testClearingAndEnteringTextIntoViewWithAccessibilityLabel, The step timed out after 10.00 seconds: Could not find element with BLOCKPREDICATE(0x6000030500f0)
  /Users/runner/work/KIF/KIF/KIF Tests/AccessibilityIdentifierTests.m:73
{
    [tester clearTextFromAndThenEnterText:@"Yo" intoViewWithAccessibilityIdentifier:@"idGreeting"];
}

  testClearingAndEnteringTextIntoViewWithAccessibilityLabel, The step timed out after 10.00 seconds: Found an accessibility element with the label "Test Suite", but not with the traits "1"
  /Users/runner/work/KIF/KIF/KIF Tests/AccessibilityIdentifierTests.m:137
{
    [tester tapViewWithAccessibilityLabel:@"Test Suite" traits:UIAccessibilityTraitButton];
}
  ```

Based on the error messages, it appears that the runner failed to enter the `TapViewController` and got stuck at the `TestSuiteViewController`. Not sure about the cause of this though.

justinseanmartin (approved):

justinseanmartin:
I think I've got a fix for the `AccessibilityIdentifierTests` flakiness in PR #1288. I only reproduced it locally on my machine once, but I'm going to run it against CI and see if it is more reliably green.

PR #1288, opened 2023-07-13
PullToRefreshTests can exit before the greetingView is finished animating out

justinseanmartin:
I was able to reproduce this issue one time on my local machine. There is a small time window where Bingo is no longer present on the screen, but the greeting view has not yet completely finished animating out. When this happens, the tap for the subsequent test `AccessibilityIdentifierTests testClearingAndEnteringTextIntoViewWithAccessibilityLabel` will fail, because it won't have successfully opened the `Tapping` tests screen.


justinseanmartin:
Green on the first attempt, so seems promising!

dostrander (approved):

PR #1289, opened 2023-10-04
iOS 17 - fix black screenshots after using a keyboard

banjun:
On iOS 17 Simulator, we cannot render a keyboard window and get a fully black screenshot image.
As a workaround, this PR suppress rendering the keyboard window

## conditions for reproduce

* call `writeScreenshotForLine` after using a keyboard
    * the keyboard window renders black pixels even when the keyboard is hidden on iOS 17 Simulator

iOS 16.4 Simulator does not reproduce the issue, for both builds on Xcode 14.3.1 or Xcode 15.0. It seems to depend on the runtime simulator version.

## expected result

Either the keyboard window is not rendered, or, it renders the keyboard image over the app window as same as displayed on the Simulator

## actual result

a fully black screenshot image will be rendered

### DIFF
--- a/Sources/KIF/Additions/UIApplication-KIFAdditions.m
+++ b/Sources/KIF/Additions/UIApplication-KIFAdditions.m
@@ -188,8 +188,10 @@ static const void *KIFRunLoopModesKey = &KIFRunLoopModesKey;
 			continue;
 		}
 
-        if (window == keyboardWindow) {
-            continue;
+        if (@available(iOS 17.0, *)) {
+            if (window == keyboardWindow) {
+                continue;
+            }
         }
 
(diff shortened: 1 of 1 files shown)


justinseanmartin:
Thanks for the contribution! The CI failures look unrelated and the change itself seems benign, given that we've never been able to render the keyboard window into the captured screenshots. I'll wait to see if anything else falls out of the Xcode 15 bump that needs to be merged before cutting a release. In the meantime, hopefully pointing at HEAD of master is fine. I'll plan to cut a release in a week or two, but feel free to poke me if I forget.

justinseanmartin:
It seems there are some reliability issues on Xcode 14.1 running the `WebViewTests testScrolling` test. I'm kicking the CI pipeline one more time. 🤞

justinseanmartin (approved):

PR #1291, opened 2024-01-18
Add support iOS Simulator 16.4+

COJIDAT:
Under iOS Simulator 16.4+ invoke [[AccessibilitySettingsController alloc] init] open lockscreen


justinseanmartin:
Is this still an issue? I haven't seen any other reports of this. I'm going to close this for now, but feel free to re-open if you thnk this is a problem still and we can discuss. CI is running green on iOS 16, 17, and 18 simulators without this change.

PR #1292, opened 2024-03-15
Bump ios minimum deployment target - New PR

adamsousa:
New PR requested by @justinseanmartin

Original PR by @danielob #1283


adamsousa:
@justinseanmartin Hmm, 3 checks were canceled almost imminently after creating the PR. Logs are unavailable to me so I'm not able to see what's going on.

justinseanmartin:
Just needed approval to run. I hit the button on this, will see what comes out of the results. Thanks for picking this up!

justinseanmartin:
Must've been flakes, because everything passed on re-run. Let me take a look at the code changes.

justinseanmartin (approved):
Thanks for the PR! The repo readme should probably say iOS 11+ instead of 11-13, but your change at least makes what we report in the podspec align with the repo readme, which is great.

PR #1295, opened 2024-07-09
Adding support for accessibility custom actions

RoyalPineapple:
Custom actions allow developers to provide custom interactions that can be initiated through assistive technologies like VoiceOver and Switch Control. For example, VoiceOver lets users access actions quickly using the Actions rotor.
Custom actions are relied on heavily by SwiftUI's `.accessibilityElement(children:.combine)`  view modifier. 

This PR introduces two methods on KIFUIViewTestActor to enable validation and activation of an element's custom actions within KIF tests. 

`- (instancetype)usingCustomActionWithName:(NSString *)name;` adds a predicate that searches elements for one that provides a custom action with a specified name. 

`- (void)activateCustomActionWithName:(NSString *)name;` first performs a search for an element that provides a custom action with a specified name and then activates the action.

### DIFF
--- a/KIF.xcodeproj/project.pbxproj
+++ b/KIF.xcodeproj/project.pbxproj
@@ -204,6 +204,11 @@
 		9CC881AC1AD4CE4B002CD34C /* Foundation.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = AAB0726B139719AC008AF393 /* Foundation.framework */; };
 		9CC881AD1AD4CE50002CD34C /* CoreFoundation.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 9CC881A21AD4CAAC002CD34C /* CoreFoundation.framework */; };
 		9CC967401AD4B1B600576D13 /* KIF.h in Headers */ = {isa = PBXBuildFile; fileRef = 9CC9673F1AD4B1B600576D13 /* KIF.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		ACA242E42C3DA55400E6F1B6 /* CustomActionTests_ViewTestActor.m in Sources */ = {isa = PBXBuildFile; fileRef = ACA242E32C3DA55400E6F1B6 /* CustomActionTests_ViewTestActor.m */; };
+		ACA242E62C3DB46A00E6F1B6 /* UIAccessibilityCustomAction+KIFAdditions.m in Sources */ = {isa = PBXBuildFile; fileRef = ACA242E52C3DB46A00E6F1B6 /* UIAccessibilityCustomAction+KIFAdditions.m */; };
+		ACA242E72C3DB47400E6F1B6 /* UIAccessibilityCustomAction+KIFAdditions.m in Sources */ = {isa = PBXBuildFile; fileRef = ACA242E52C3DB46A00E6F1B6 /* UIAccessibilityCustomAction+KIFAdditions.m */; };
+		ACA242E92C3DB4EA00E6F1B6 /* UIAccessibilityCustomAction+KIFAdditions.h in Headers */ = {isa = PBXBuildFile; fileRef = ACA242E82C3DB4B000E6F1B6 /* UIAccessibilityCustomAction+KIFAdditions.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		ACA242EA2C3DB4EB00E6F1B6 /* UIAccessibilityCustomAction+KIFAdditions.h in Headers */ = {isa = PBXBuildFile; fileRef = ACA242E82C3DB4B000E6F1B6 /* UIAccessibilityCustomAction+KIFAdditions.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		AE62FCD01A1D20E5002B10DA /* WebViewTests.m in Sources */ = {isa = PBXBuildFile; fileRef = AE62FCCF1A1D20E5002B10DA /* WebViewTests.m */; };
 		AE62FCD61A1D2447002B10DA /* WebViewController.m in Sources */ = {isa = PBXBuildFile; fileRef = AE62FCD51A1D2447002B10DA /* WebViewController.m */; };
 		AE62FCD81A1D2667002B10DA /* index.html in Resources */ = {isa = PBXBuildFile; fileRef = AE62FCD71A1D2667002B10DA /* index.html */; };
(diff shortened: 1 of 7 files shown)


Sources/KIF/Additions/UIAccessibilityCustomAction+KIFAdditions.m
```diff
@@ -0,0 +1,46 @@
...
+
+#import <UIKit/UIKit.h>
+#import "UIAccessibilityCustomAction+KIFAdditions.h"
+
+@implementation UIAccessibilityCustomAction (KIFAdditions)
+
+- (BOOL)activate;
+{
+    if (@available(iOS 13.0, *)) {
+        if (self.actionHandler && self.actionHandler(self)) {
```
kyleve, line 16: Assuming it's intentional we don't return the result of `actionHandler` here?

kyleve (approved):

Sources/KIF/Additions/UIAccessibilityCustomAction+KIFAdditions.m
RoyalPineapple: the docs say that if theres an action handler it takes precedence, but nothing prevents a UIAccessibilityCustomAction from having both an actionHandler and a target/selector pair. 

This allows us to give the target/selector (if it exists) an opportunity to succeed (return true) if the actionHandler doesn't..

RoyalPineapple:
@justinseanmartin What do you think of this proposed API? We're going to start relying on Custom Actions much more frequently.

Sources/KIF/Classes/KIFUIViewTestActor.m
```diff
@@ -155,6 +156,26 @@ - (instancetype)usingFirstResponder;
     return [self usingPredicate:predicate];
 }
 
+- (instancetype)usingCustomActionWithName:(NSString *)name
+{
+    NSPredicate *predicate = [NSPredicate predicateWithBlock:^BOOL(id evaluatedObject, NSDictionary *bindings) {
+        NSArray *actions = [evaluatedObject accessibilityCustomActions];
```
justinseanmartin, line 162: Every element that we iterate over in the view & accessibility element hierarchy is guaranteed to respond to this selector? Or should we add a `respondsToSelector:` check?

Sources/KIF/Classes/KIFUIViewTestActor.m
```diff
@@ -155,6 +156,26 @@ - (instancetype)usingFirstResponder;
...
+        for (UIAccessibilityCustomAction *action in actions) {
+            NSString *actionName = [action name];
+            if ([actionName isKindOfClass:[NSAttributedString class]]) {
+                actionName = [(NSAttributedString *)actionName string];
+            }
+
+            if ([actionName isEqualToString: name]) {
+                return true;
+            }
+        }
```
justinseanmartin, line 172: NIT: Up to you, but maybe worth a helper method for getting the specified action by name? Its identical here and below and might make these methods slightly simpler.

Sources/KIF/Classes/KIFUIViewTestActor.m
```diff
@@ -385,6 +406,31 @@ - (void)swipeFromEdge:(UIRectEdge)edge
     [self.actor swipeFromEdge:edge];
 }
 
+- (void)activateCustomActionWithName:(NSString *)name;
+{
+    KIFUIObject *found = [[self usingCustomActionWithName:name] _predicateSearchWithRequiresMatch:YES mustBeTappable:NO];
```
justinseanmartin, line 411: We generally wrap the body of these methods that return a `KIFUIObject` in an `@autoreleasepool` block so as to not accidentally extend the lifetime of the view object beyond this action.

Sources/KIF/Classes/KIFUIViewTestActor.m
```diff
@@ -385,6 +406,31 @@ - (void)swipeFromEdge:(UIRectEdge)edge
...
+    
+    NSArray *actions = [found.element accessibilityCustomActions];
```
justinseanmartin, line 413: NIT: Any reason to do this outside the block vs inside it? Don't feel strongly, just not what I would've expected. This is potentially moot if we refactor to using a helper method like `found.element customActionWithName:name` or something like that.

Sources/KIF/Additions/UIAccessibilityCustomAction+KIFAdditions.h
```diff
@@ -0,0 +1,15 @@
...
+//  KIF
+//
+//  Created by Alex Odawa on 09/07/2024.
+//
+
+#import <UIKit/UIKit.h>
+
+@interface UIAccessibilityCustomAction (KIFAdditions)
+
+- (BOOL)activate;
```
justinseanmartin, line 12: Use a prefix to avoid collisions, e.g. `KIF_activate`

Sources/KIF/Additions/UIAccessibilityCustomAction+KIFAdditions.m
```diff
@@ -0,0 +1,46 @@
...
+            id arg = self;
+            [invocation setArgument: &arg atIndex:2];
+        }
+        
+        [invocation invoke];
+        BOOL returnValue = NO;
+        [invocation getReturnValue:&returnValue];
+        return returnValue;
+    }
+    return NO;
```
justinseanmartin, line 43: Would it make more sense to raise if we attempted to perform an action on an element that says it supports it, but hasn't implemented the selector? I'd kind of expect this to crash or raise an error in some way that is distinct from the selector returning false from the action.

Sources/KIF/Classes/KIFUIViewTestActor.m
```diff
@@ -385,6 +406,31 @@ - (void)swipeFromEdge:(UIRectEdge)edge
...
+    
+    [self runBlock:^KIFTestStepResult(NSError **error) {
+        for (UIAccessibilityCustomAction *action in actions) {
+            NSString *actionName = [action name];
+            if ([actionName isKindOfClass:[NSAttributedString class]]) {
+                actionName = [(NSAttributedString *)actionName string];
+            }
+
+            if ([actionName isEqualToString: name]) {
+                if([action activate]) {
```
justinseanmartin, line 423: How do VoiceOver and other accessibility tools respond if a custom action returns false? Are there valid use cases where a custom action might want to return false and that be a valid test scenario rather than treating it as a test failure? We don't need to extend to support that right now, but it seemed worth some thought/discussion. 

Is it reasonable that a custom action might fail in some cases? Or is the expectation that custom actions should generally only be present on an element if they're expected to be able to be performed? Looking at the [Firefox iOS codebase](https://github.com/mozilla-mobile/firefox-ios/blob/d8be21cb8b7ec11087138f11166883870daa6571/firefox-ios/Client/Frontend/Browser/BrowserViewController/Views/BrowserViewController.swift#L726) as an example, it seems like they filter the actions to ones that are expected to be usable in the current context, and then always return true. That's just one example though.

justinseanmartin (commented):
Mostly NITs or questions, but I think there probably are some minor changes to be made at least for prefixing the method in the category, the autoreleasepool and potentially the respondsToSelector check.

Sources/KIF/Classes/KIFUIViewTestActor.m
RoyalPineapple: This is a property on NSObject defined thusly in UIAccessibility.h
Though a respondsToSelector: check couldn't hurt.

`@property (nullable, nonatomic, strong) NSArray <UIAccessibilityCustomAction *> *accessibilityCustomActions API_AVAILABLE(ios(8.0));`

Sources/KIF/Additions/UIAccessibilityCustomAction+KIFAdditions.h
RoyalPineapple: oh man, i honestly do love objective c

Sources/KIF/Classes/KIFUIViewTestActor.m
RoyalPineapple: theres no official dogma on this particular question, but as we interpret the default behavior of UIKit and SwiftUI as guidance it would suggest that custom actions that are known to fail should be omitted entirely. 

To the person in front of the device, a failed action will play an additional "error" sound when activated. 

You're right, we should consider the case where an action is expected to occasionally fail and allow for testing that assumption.

Sources/KIF/Additions/UIAccessibilityCustomAction+KIFAdditions.m
RoyalPineapple: Ohh, nice idea

Sources/KIF/Classes/KIFUIViewTestActor.m
RoyalPineapple: mooted

KIF.xcodeproj/project.pbxproj
```diff
@@ -1805,7 +1805,7 @@
 				GCC_WARN_UNINITIALIZED_AUTOS = YES;
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
-				IPHONEOS_DEPLOYMENT_TARGET = 11.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 12.0;
```
RoyalPineapple, line 1808: This silences a warning

Sources/KIF/Additions/UIView-Debugging.m
```diff
@@ -136,6 +136,9 @@ - (void)printDatePickerState {
         case UIDatePickerModeCountDownTimer:
             printf(" UIDatePickerModeCountDownTimer");
             break;
+        case UIDatePickerModeYearAndMonth:
```
RoyalPineapple, line 139: This silences an error

RoyalPineapple: This needs to be set for the spec to validate on new xcode, but also needs to be excluded for the tests to pass on older versions. fun

PR #1296, opened 2024-07-12
Introducing tryFindingViewInFrame() to avoid scroll

SimoncelloCT:
# Description
This MR introduces the method `tryFindingViewInFrame(withAccessibilityIdentifier: )` that allows to check the visibility of an element **in the current frame** as it is showed to the user, without any additional interaction.
The method usingCurrentFrame() has been added in the new `KIFUIViewTestActor` to maintain parity with `KIFUITestActor`.

### Why?
In some UITests we want to ensure an element is visible in the **current frame**. 

When the element we want to assert is visible is inside a `scrollView`,`tableView` or `collectionView` the method `tryFindingView(withAccessibilityIdentifier: )` causes a scroll to the offset where the element is placed, making it visible even if it wasn't. This behaviour does not allow to have reliable visibility checks (eg. if an element exists but is not actually visible, the method returns true).

**EDIT:** this MR also fixes a bug: under some conditions a control was considered `tappable` when just `enabled` and having a `TapGestureRecognizer`, even when it was occluded by another element.

### DIFF
--- a/Sources/KIF/Additions/UIAccessibilityElement-KIFAdditions.m
+++ b/Sources/KIF/Additions/UIAccessibilityElement-KIFAdditions.m
@@ -99,7 +99,7 @@ MAKE_CATEGORIES_LOADABLE(UIAccessibilityElement_KIFAdditions)
 
     if (!element) {
         if (error) {
-            *error = [self errorForFailingPredicate:predicate];
+            *error = [self errorForFailingPredicate:predicate disableScroll:scrollDisabled];
         }
         return NO;
     }
@@ -276,9 +276,9 @@ MAKE_CATEGORIES_LOADABLE(UIAccessibilityElement_KIFAdditions)
     return view;
 }
 
-+ (NSError *)errorForFailingPredicate:(NSPredicate*)failingPredicate;
++ (NSError *)errorForFailingPredicate:(NSPredicate*)failingPredicate disableScroll:(BOOL) scrollDisabled;
 {
-    NSPredicate *closestMatchingPredicate = [self findClosestMatchingPredicate:failingPredicate];
+    NSPredicate *closestMatchingPredicate = [self findClosestMatchingPredicate:failingPredicate disableScroll:scrollDisabled];
     if (closestMatchingPredicate) {
         return [NSError KIFErrorWithFormat:@"Found element with %@ but not %@", \
                 closestMatchingPredicate.kifPredicateDescription, \
@@ -287,7 +287,7 @@ MAKE_CATEGORIES_LOADABLE(UIAccessibilityElement_KIFAdditions)
     return [NSError KIFErrorWithFormat:@"Could not find element with %@", failingPredicate.kifPredicateDescription];
 }
 
-+ (NSPredicate *)findClosestMatchingPredicate:(NSPredicate *)aPredicate;
++ (NSPredicate *)findClosestMatchingPredicate:(NSPredicate *)aPredicate disableScroll:(BOOL) scrollDisabled;
 {
     if (!aPredicate) {
         return nil;
@@ -295,7 +295,7 @@ MAKE_CATEGORIES_LOADABLE(UIAccessibilityElement_KIFAdditions)
     
     UIAccessibilityElement *match = [[UIApplication sharedApplication] accessibilityElementMatchingBlock:^BOOL (UIAccessibilityElement *element) {
         return [aPredicate evaluateWithObject:element];
-    }];
+    } disableScroll:scrollDisabled];
     if (match) {
         return aPredicate;
     }
@@ -313,7 +313,7 @@ MAKE_CATEGORIES_LOADABLE(UIAccessibilityElement_KIFAdditions)
             if (predicateMinusOneCondition) {
                 UIAccessibilityElement *match = [[UIApplication sharedApplication] accessibilityElementMatchingBlock:^BOOL (UIAccessibilityElement *element) {
                     return [predicateMinusOneCondition evaluateWithObject:element];
-                }];
+                } disableScroll:scrollDisabled];
                 if (match) {
                     return predicateMinusOneCondition;
                 }
--- a/Sources/KIF/Additions/UIAccessibilityElement-KIFAdditions.m
+++ b/Sources/KIF/Additions/UIAccessibilityElement-KIFAdditions.m
@@ -59,12 +59,17 @@ MAKE_CATEGORIES_LOADABLE(UIAccessibilityElement_KIFAdditions)
 
 + (BOOL)accessibilityElement:(out UIAccessibilityElement **)foundElement view:(out UIView **)foundView withLabel:(NSString *)label value:(NSString *)value traits:(UIAccessibilityTraits)traits fromRootView:(UIView *)fromView tappable:(BOOL)mustBeTappable error:(out NSError **)error
 {
+    return [self accessibilityElement:foundElement view:foundView withLabel:label value:value traits:traits fromRootView:fromView tappable:mustBeTappable error:error disableScroll:NO];
+}
+
++ (BOOL)accessibilityElement:(out UIAccessibilityElement **)foundElement view:(out UIView **)foundView withLabel:(NSString *)label value:(NSString *)value traits:(UIAccessibilityTraits)traits fromRootView:(UIView *)fromView tappable:(BOOL)mustBeTappable error:(out NSError **)error disableScroll:(BOOL)scrollDisabled
+{
     UIAccessibilityElement *element = [self accessibilityElementWithLabel:label value:value traits:traits fromRootView:fromView error:error];
     if (!element) {
         return NO;
     }
     
-    UIView *view = [self viewContainingAccessibilityElement:element tappable:mustBeTappable error:error];
+    UIView *view = [self viewContainingAccessibilityElement:element tappable:mustBeTappable error:error disableScroll:scrollDisabled];
     if (!view) {
         return NO;
     }
@@ -83,10 +88,15 @@ MAKE_CATEGORIES_LOADABLE(UIAccessibilityElement_KIFAdditions)
 
 + (BOOL)accessibilityElement:(out UIAccessibilityElement **)foundElement view:(out UIView **)foundView withElementMatchingPredicate:(NSPredicate *)predicate tappable:(BOOL)mustBeTappable error:(out NSError **)error;
 {
+    return [self accessibilityElement:foundElement view:foundView withElementMatchingPredicate:predicate tappable:mustBeTappable error:error disableScroll:NO];
+}
+
++ (BOOL)accessibilityElement:(out UIAccessibilityElement **)foundElement view:(out UIView **)foundView withElementMatchingPredicate:(NSPredicate *)predicate tappable:(BOOL)mustBeTappable error:(out NSError **)error disableScroll:(BOOL)scrollDisabled;
+{
     UIAccessibilityElement *element = [[UIApplication sharedApplication] accessibilityElementMatchingBlock:^BOOL(UIAccessibilityElement *element) {
         return [predicate evaluateWithObject:element];
-    }];
-    
+    } disableScroll: scrollDisabled];
+
     if (!element) {
         if (error) {
             *error = [self errorForFailingPredicate:predicate];
@@ -94,7 +104,7 @@ MAKE_CATEGORIES_LOADABLE(UIAccessibilityElement_KIFAdditions)
         return NO;
     }
     
-    UIView *view = [UIAccessibilityElement viewContainingAccessibilityElement:element tappable:mustBeTappable error:error];
+    UIView *view = [UIAccessibilityElement viewContainingAccessibilityElement:element tappable:mustBeTappable error:error disableScroll:scrollDisabled];
     if (!view) {
         return NO;
     }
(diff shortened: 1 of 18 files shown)


Sources/KIF/IdentifierTests/KIFUITestActor-IdentifierTests.h
```diff
@@ -85,6 +85,13 @@
  */
 - (BOOL) tryFindingViewWithAccessibilityIdentifier:(NSString *) accessibilityIdentifier;
 
+/*!
+ @abstract returns YES or NO if the element is visible in the current frame (without scrolling).
+ @discussion if the element described by the accessibility identifier is visible, the method returns true.
+ @param accessibilityIdentifier The accessibility identifier of the element to query for
+ */
+- (BOOL) tryFindingViewInFrameWithAccessibilityIdentifier:(NSString *) accessibilityIdentifier;
```
justinseanmartin, line 93: We've been trying to maintain parity between `KIFUITestActor` (the old API) and `KIFUIViewTestActor` (the new API). Would you be willing to add a method like `usingCurrentFrame` or `usingAutomaticScrolling:NO` to `KIFUIViewTestActor`. This would be similar to how `usingTimeout:` works on KIFTestActor to alter matching behavior, but not necessarily altering the predicates used to do the matching. It'd store that state in the actor (defaults to scrolling if not specified) and then pass the scrolling behavior property through [here](https://github.com/kif-framework/KIF/blob/master/Sources/KIF/Classes/KIFUIViewTestActor.m#L610) (and a few lines below). Thoughts?

justinseanmartin (commented):
Thank you so much for the contribution! Would you be able to add an integration test that verifies this functionality is working as expected and ensures we don't regress it going forward?

Sources/KIF/IdentifierTests/KIFUITestActor-IdentifierTests.h
SimoncelloCT: Hey @justinseanmartin, yes it totally makes sense! I'm going to commit these changes today 👌

SimoncelloCT:
Hey @justinseanmartin , I added all the changes we've discussed and also fixed somethings specific for `scrollViews`, I found my prev solution was working only with `TableViews` and `CollectionViews`. 

I checked that all the tests are passing as expected: 
Screenshot 2024-07-24 at 12 51 06

Thank you for the guidance and hope to see this PR merged soon 😄  🙏

Sources/KIF/IdentifierTests/KIFUITestActor-IdentifierTests.h
SimoncelloCT: Take a look now 😄

justinseanmartin: Looks great, thanks for the iteration!

justinseanmartin (approved):

justinseanmartin:
I've kicked off a CI run and will check back on it in the morning to see if it is good to merge. There have been a couple of flaky tests in the integration test suite, so I'll rekick the job once or twice if it looks like that's happening.

Sources/KIF/Additions/UIView-Debugging.m
```diff
@@ -136,6 +136,9 @@ - (void)printDatePickerState {
         case UIDatePickerModeCountDownTimer:
             printf(" UIDatePickerModeCountDownTimer");
             break;
+        case UIDatePickerModeYearAndMonth:
```
SimoncelloCT, line 139: I see we have a failure on CI due to this addition. It is not related to my functionality, I added it to make it compile while running tests locally, so it is weird 🤔
Should I remove it?

justinseanmartin: Sorry, didn't get a chance to circle back around to this today. If I don't get time this weekend I'll set a reminder for Monday morning to check out the failure.

If you know it is related to the code you added and is tangential to the fix, it's up to you if you think the best path is removing it. I'll have to take a closer look.

Sources/KIF/Additions/UIView-KIFAdditions.m
```diff
@@ -873,8 +873,8 @@ - (BOOL)isProbablyTappable
 // Is this view currently on screen?
 - (BOOL)isTappable;
 {
-    return ([self hasTapGestureRecognizerAndIsControlEnabled] ||
-            [self isTappableInRect:self.bounds]);
+    return ([self isTappableInRect:self.bounds] || ([self hasTapGestureRecognizerAndIsControlEnabled] &&
+            [self isTappableInRect:self.bounds]));
```
SimoncelloCT, line 877: Hello @justinseanmartin , I've found a bug here: a view was considered tappable even when occluded by another element, because `isTappable` was returning `true` even when only `hasTapGestureRecognizerAndIsControlEnabled` was true. 

Since this is a reliability issue, I thought to take advantage of the MR that has not been merged yet and fix it. I've added also integration tests to make sure it will be maintained over time 😀

I edited also the MR description, let me know if something is not clear 🙏🏻

justinseanmartin: Was this breaking something for you, or just something you noticed that seemed wrong while you were in here?

I want to say that this was by design, because sometimes the tap event might trigger even if the element isn't the one directly that gets returned by a hit test. I'll verify by pointing our internal tests at this branch and see if anything breaks.

IIUC - I think the boolean logic you have here is equivalent to `return [self isTappableInRect:self.bounds];`, disregarding whether `hasTapGestureRecognizerAndIsControlEnabled` is true or false. If that's the case, we should simplify this, but I do think there are cases where this was necessary.

SimoncelloCT: Thanks for your answer! Indeed the boolean logic is the same, it can be simplified if we leave this change (sorry for that, just a bit tired 😅).
Yes it breaks `tappable` checks: calling `tryFindingTappableView()` I was getting `true` for occluded views. 
You can easily check it running the integration test I've added `testTryFindingOccludedTappableViewInFrameWithAccessibilityIdentifier` without this change. 

If you have a better idea on how to fix it, let me know and I can revert this change and follow your guidance 👍

justinseanmartin: Looking at the history, we tried to remove this check in #1244 and ran into problems that led us to reverting that change in #1250.

I think the way I'd probably address this for now would be in `isTappable` to grab the intersection of the `self.bounds` rect with the view's window (might require translating coordinate systems) that pass that into `isTappableInRect:`. We could probably add a small optimization in `tappablePointInRect:` to early return with NO if the supplied rect has 0 by 0 dimensions.

justinseanmartin: I confirmed that it does break a couple tests on a few screens in our app, but relatively minimal breakage. It might be something that we can work around on our end. In particular, I saw it trying to tap on a search field's placeholder text and saying it wasn't tappable. If we can avoid the breakage while still getting you the functionality that you need, that seems preferable.

SimoncelloCT: Ok, I'll try to follow your idea and commit the changes today, I'll keep you posted

justinseanmartin:
FYI - you might want to merge in or rebase on top of https://github.com/kif-framework/KIF/pull/1297

Sources/KIF/Additions/UIView-KIFAdditions.m
justinseanmartin: Sorry, I'm still digging into this to see if there is something that we can do on our end to work around it. This is in a custom search field control using [Blueprint](https://github.com/square/Blueprint). Given that it is only happening in one place in our codebase, it might be a problem with which view the accessibility identifier is being set on. Unfortunately I've got limited time today and am traveling the rest of the week, so it might be a little slow going on my end, but I will get back to looking at this next week.

If you can maintain the existing behavior regarding the gesture recognizers and only exclude offscreen elements in the tappability check, I think that'd be less likely to cause issues for existing tests. If not, no worries, just might take a little longer to dig through this on my end. In the meantime, you can depend on your unpublished KIF fork via git/branch to unblock yourself in the meantime and we'll get something that works here landed in the near future.

justinseanmartin:
I think if you rebase, we should be good to land this. We're going to work to drop Xcode 13 and getting things green on Xcode 15 in #1298, but I wouldn't block this PR on that.

Sources/KIF/Additions/UIView-KIFAdditions.m
justinseanmartin: Can we add a condition to check the [isVisibleInWindowFrame](https://github.com/kif-framework/KIF/blob/8757ba89034d7a590e383558908a44cc61c702bf/Sources/KIF/Additions/UIView-KIFAdditions.m#L1072) of this view if it isn't tappable but has a gesture recognizer?

SimoncelloCT: Hey Justin, thank you for your answers, sorry but I didn't have the time to check them before.

I've tried with a check like: 
```
([self isTappableInRect:self.bounds] || ([self hasTapGestureRecognizerAndIsControlEnabled] &&
            [self isVisibleInWindowFrame]));
```
but we have the same issue: An element that has a tap gesture recognizer and is visible in the current frame **but occluded by another view** is considered as tappable and this breaks some tests on our side.. 

So we still have the same issue, the element is actually not tappable (because totally occluded) but considered as it is 😢
I don't think there is a way to consider an element tappable when the check `isTappableInRect:self.bounds` fails, without affecting the reliability of this method tbh, but I might still miss something

What do you think?

SimoncelloCT:
@justinseanmartin  The branch has been rebased on top of master, let me know if we could still try something for the tappability check or it is good to go!  🙏🏻

Sources/KIF/Additions/UIView-KIFAdditions.m
justinseanmartin: Ok, I figured out an issues with one of our test suites caused by this change.

We have a search bar element that contains a search text field as a subview. The text field is being marked with `userInteractionEnabled = false`, because the tap handler for the search bar redirects to another screen instead of allowing to type into it directly the text field. However, the accessibilityIdentifier we were matching on is being set on the search text field (matching what is shown in the text field placeholder). This works today because the element matching the hitTest (the search bar) has a tap gesture recognizer on it.

In our case, I can change the predicate to match on the accessibilityIdentifier of the search bar instead and things seem to work in this case. That said, this would also potentially break tests at other companies as well. I also still have another failure to look into still that isn't related to this case.

On the surface, it seems like it should be reasonable to allow the tap action to be handled by a superview that can process touch events. From a users perspective, they don't know or care if the tap is being handled by the search text field or the search bar. That said, trying to make `isTappableWithHitTestResultView:` allow for this seems to cause different problems, where we'll match on occluded elements that don't actually respond to touch events (like a label within a button).

I'm still thinking about this, but wanted to give a bit of an update here.

SimoncelloCT: Yeah, I got it, thank you for the update. We can keep each other updated here, I'll let you know if any idea comes to my mind when I'll have the time to work a bit on this

Sources/KIF/Additions/UIAccessibilityElement-KIFAdditions.m
```diff
@@ -105,19 +115,24 @@ + (BOOL)accessibilityElement:(out UIAccessibilityElement **)foundElement view:(o
 }
 
 + (BOOL)accessibilityElement:(out UIAccessibilityElement *__autoreleasing *)foundElement view:(out UIView *__autoreleasing *)foundView withElementMatchingPredicate:(NSPredicate *)predicate fromRootView:(UIView *)fromView tappable:(BOOL)mustBeTappable error:(out NSError *__autoreleasing *)error
+{
+    return [self accessibilityElement: foundElement view:foundView withElementMatchingPredicate:predicate tappable:mustBeTappable error:error disableScroll:NO];
```
justinseanmartin, line 119: Ok, I found the bug leading to the other set of tests that were failing in our repo, and it is caused by dropping this param here:

```suggestion
    return [self accessibilityElement:foundElement view:foundView withElementMatchingPredicate:predicate fromRootView:fromView tappable:mustBeTappable error:error disableScroll:NO];
```

SimoncelloCT: Sorry for that, it is just a mistake from my side. I've applied your suggestion! 🙏🏻

SimoncelloCT:
I see 2 tests were broken because of a duplicate label in the `TextField` I've added for the new tests, I've pushed a fix for that

Sources/KIF/Additions/UIView-KIFAdditions.m
SimoncelloCT: Hey @justinseanmartin, I've experimented a bit more too but I wasn't able to find a solution to allow both cases 😞. 

_On a more personal note, I think it's reasonable in UITests to refer always to the actual element that should handle the tap (this is what we usually do, but as you mentioned I don't know if referring to a subview, which cannot handle tap events, is something usual in in other companies tests. The change might be justified in that case if we confirm there is no better solution to make the check reliable)._

Are you still trying to find a solution for this?

I'm sorry to push for an answer, the fact is that if we cannot merge the change for the reason above we will create a fork in our company GitHub account to avoid keeping reference to my personal account's fork
Thanks 🙏

justinseanmartin:
> @justinseanmartin I see 2 tests were broken because of a duplicate label in the `TextField` I've added for the new tests, I've pushed a fix for that, now all tests should pass

@SimoncelloCT - Hey, apologies for the delay here. I'm working on fixing the CI configuration and getting it to run on Xcode 14 & 15 successfully. Once that's done, we should be able to land this. I also found a problem in the `pullToRefresh` tests that would could help to use your new `usingCurrentFrame`. I'm actively working on it this week, so hopefully should have this in a state to be landed soon.

justinseanmartin:
I've got the CI fixes in #1300. I've confirmed that this works in that context, so I'm going to land this PR, even though it has red builds.

PR #1300, opened 2024-08-08
Get KIF CI working reliably with Xcode 14.3.1 & 15.4

justinseanmartin:
I'd recommend reviewing this commit-by-commit, as they're logically organized and have verbose commit messages. Some notable things:

**KIF behavior changes:**
* 424ed5fdaee10fea76914c0e3e0c3608bf954ce9: Increases the KIF animation wait time from .5s to 2s by default
* 0dca5ba26a2fd126d983f247f70d2b0cbf72ab19: Slightly alters the pullToRefresh logic to be more realistic to match a users behavior.
  * I could be convinced to drop this commit. It was somewhat ancillary to the rest of the changes in this PR. I thought it was related to the PullToRefresh tests being flaky, but in reality that was mostly caused by the changes in #1020.

**PullToRefresh test flakiness:**
This was probably the trickiest thing that I discovered while digging in here. There were a bunch of flakes happening on CI that I couldn't really explain. What I saw when running these tests locally was the pull to refresh control would be dismissed abruptly after the action completed, despite it being configured to wait for up to 4s to complete. What I found was that the behavior in #1020 to instantly scroll without animations to enumerate table view cells is what is causing the abrupt cancellation of the pull to refresh operation. This was unexpected and I think a side-effect of the behavior change in #1020. 

Luckily, in #1296, @SimoncelloCT added the capability to prevent scrolling the view when enumerating the view hierarchy. I leveraged the functionality there to stabilize these tests.

**Other test flakiness:**
There were a bunch of small changes required to get the tests to pass more reliably. I tried reproducing a lot of these with slow animations enabled locally in the simulator settings. CI is working much better now with this change.

**Misc cleanup:**
* I got rid of Carthage references that weren't being validated. I'm not sure if it still works or not, but we weren't validating it, so I'm just reflecting that state of the world. 
* The KIFFramework target (and respective tests) target wasn't compiling and I couldn't find an easy way to fix that and not break SPM at the same time. For the TestableSwift and Calculator example projects, had them consume KIF via SPM instead.
* I got rid of spaces from the folders and reorganized the files in the file system to make it cleaner. The top level folder now looks like this:
image


congt (approved):
Pair reviewed with @justinseanmartin offline

PR #1304, opened 2024-09-17
added performAccessibilityActivateWithExpectedResult

RoyalPineapple:
Some of my recent accessibility work in our design system results in accessibility elements that abstract away the interactivity of some underlying controls. These elements subsume the interactivity of their children and re-expose the functionality via `accessibilityActivate()`.

Unfortunately this breaks KIF matching resulting in the need for an escape hatch to allow activation of elements with greater complexity.

performAccessibilityActivateWithExpectedResult calls accessibilityElement() on the found element and compares its result to the expectation.

### DIFF
--- a/Sources/KIF/Classes/KIFUIViewTestActor.h
+++ b/Sources/KIF/Classes/KIFUIViewTestActor.h
@@ -200,6 +200,11 @@ extern NSString *const inputFieldTestString;
  */
 - (void)activateCustomActionWithName:(NSString *)name expectedResult:(BOOL)expectedResult;
 
+/*!
+ @abstract Activates a found element via `accessibilityActivate()`.
+ @param expectedResult The expected boolean return from activation of the element.
+ */
+- (void)performAccessibilityActivateWithExpectedResult:(BOOL)expectedResult;
 
 #pragma mark Waiting & Finding
 
--- a/Sources/KIF/Classes/KIFUIViewTestActor.m
+++ b/Sources/KIF/Classes/KIFUIViewTestActor.m
@@ -403,6 +403,8 @@ NSString *const inputFieldTestString = @"Testing";
     [self.actor swipeFromEdge:edge];
 }
 
+#pragma mark - Accesibility Actions
+
 - (void)activateCustomActionWithName:(NSString *)name;{
     [self activateCustomActionWithName:name expectedResult:YES];
 }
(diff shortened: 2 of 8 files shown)


Sources/KIF/Classes/KIFUIViewTestActor.h
```diff
@@ -192,6 +192,11 @@ extern NSString *const inputFieldTestString;
  */
 - (void)activateCustomActionWithName:(NSString *)name expectedResult:(BOOL)expectedResult;
 
+/*!
+ @abstract Activates a found element via `accessibilityActivate()`..
```
RoyalPineapple, line 196: nit: two periods

Sources/KIF/Classes/KIFUIViewTestActor.m
```diff
@@ -411,7 +411,20 @@ - (void)activateCustomActionWithName:(NSString *)name expectedResult:(BOOL)expec
             return KIFTestStepResultFailure;
         }];
     }
+}
 
+- (void)performAccessibilityActivateWithExpectedResult:(BOOL)expectedResult;
+{
+    @autoreleasepool {
+        KIFUIObject *found = [self _predicateSearchWithRequiresMatch:YES mustBeTappable:YES];
```
justinseanmartin, line 419: Does `mustBeTappable` need to be true on these objects? I suppose you still want to limit this to objects that would presumably pass a hitTest check?

KIF Tests/AccessibilityActivationTests_ViewTestActor.m
```diff
@@ -0,0 +1,32 @@
...
+
+- (void)afterEach
+{
+    [[[viewTester usingLabel:@"Test Suite"] usingTraits:UIAccessibilityTraitButton] tap];
+}
+
+- (void)testAccessibilityActivate
+{
+    [[viewTester usingLabel:@"AccessibilityView"] performAccessibilityActivateWithExpectedResult:YES];
+    [[viewTester usingValue:@"Activated"] tryFindingView];
```
justinseanmartin, line 29: I think maybe we want `waitForView` here if it might need some time to handle the accessibility activation event. The `tryFindingView` method will assert the state immediately without any additional runloop spins.

Sources/KIF/Classes/KIFUIViewTestActor.m
```diff
@@ -411,7 +411,20 @@ - (void)activateCustomActionWithName:(NSString *)name expectedResult:(BOOL)expec
...
+        
+        [self runBlock:^KIFTestStepResult(NSError **error) {
+            if([found.element accessibilityActivate] == expectedResult) {
+                return KIFTestStepResultSuccess;
```
justinseanmartin, line 423: Should we wait for animations here before yielding back to the test? I'm presuming that `accessibilityActivate` is generally an approximation of a user action that might alter app state and cause transitions to happen?

justinseanmartin (approved):

Sources/KIF/Classes/KIFUIViewTestActor.m
RoyalPineapple: In the cases I'm imagining they'd be tappable yes. 
I'm struggling to imagine a non tappable element that would be activated, interactive elements tend to be.

Sources/KIF/Classes/KIFUIViewTestActor.m
RoyalPineapple: thats a good idea

KIF Tests/AccessibilityActivationTests_ViewTestActor.m
RoyalPineapple: smart thinking, added

Sources/KIF/Classes/KIFUIViewTestActor.m
RoyalPineapple: Came back around on this and i dont think this needs to be tappable.

kyleve (approved):

PR #1305, opened 2024-09-19
Use accessibilityActivationPoint when determining where to tap elements.

RoyalPineapple:
If a call to `accessibilityActivate()` returns false, VoiceOver (and related assistive technologies) will issue a tap event to the element as a secondary attempt to activate it via the the standard touch pipeline. The `accessibilityActivationPoint` can be set on elements to instruct assistive technologies where specifically on the screen to issue this tap event. 

We should take this property into consideration when deciding the same. This is particularly important when the value is set deliberately. Failing to do so can cause taps to be issued that don't properly activate the element. 

In the case where the property has not been explicitly set. The default value is the midpoint of the accessibility frame, which itself has a default value equivalent to self.frame this makes the change functionally equivalent to the following line which returns the midpoint of the frame.

### DIFF
--- a/Sources/KIF/Classes/KIFUITestActor.m
+++ b/Sources/KIF/Classes/KIFUITestActor.m
@@ -1673,6 +1673,14 @@ static BOOL KIFUITestActorAnimationsEnabled = YES;
 
 - (CGPoint) tappablePointInElement:(UIAccessibilityElement *)element andView:(UIView *)view
 {
+    // AccessibilityActivationPoint indicates where on the element assistive technologies should issue a tap event to activate the element.
+    // In the case where the property has not been explicitly set. The default value is the midpoint of the accessibility frame.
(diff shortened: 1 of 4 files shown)


kyleve (approved):

Sources/KIF/Additions/UIView-KIFAdditions.m
```diff
@@ -931,6 +931,13 @@ - (CGPoint)tappablePointInRect:(CGRect)rect;
     UIView *hitView = nil;
     CGPoint tapPoint = CGPointZero;
     
+    // accessibilityActivationPoint
```
justinseanmartin, line 934: Would be good to document here that it defaults to the center if no override is set.

justinseanmartin (approved):
To verify the behavior and prevent regressions, it would be nice to have a test where tapping on a control updates a label with what coordinates it was tapped at and then set a custom accessibility tap point on it.

PR #1306, opened 2024-09-23
Xcode 16 in CI

justinseanmartin:
Configure KIF CI to run on Xcode 16.0 & iOS 18.0, bringing us to a total of 6 CI configurations. I've moved the one that does the extra validation to the Xcode 16 build part. If I could test iOS 16 w/ Xcode 15, I'd drop support for Xcode 14.3.1, but that isn't set up on GH hosted runners and I think it is worth still testing on iOS 16.

Aside from needing to land the fix for UICollectionView asserts in #1307, `TypingTests testEnteringTextIntoFirstResponder` was also failing on CI and required some behavior changes to get them working reliably. The behavior was that we were attempting to tap "Select All" after a long press, but we weren't waiting long enough for the popover control to appear. Tapping too early was then dismissing the popover rather than actually performing a "Select All" operation.

The KIF framework functional changes are:
* Ensure that all windows (including the UITextEffectsWindow) get layer speed multipliers applied
  * I noticed that the text effects window wasn't matching the speed of the main window when increasing the animation speed. This is mostly a performance optimization, but it also causes issues because we use the layer speed as a scale multiplier for the animation stabilization time.
* Improve detection of in progress animations
  * The test wasn't correctly waiting until the popover control fully appeared with "Slow Animations" setting enabled on the iOS simulator. They also wouldn't behave correctly with the layer speed cranked up to 100x. The change in this commit works more reliably, though it does rely on some non-public Apple APIs.
* Set .25s min for animation stabilization wait time, regardless of layer speed
  * When the animation layer speed is increased to 100x, there is no animation used for presenting the text popover. This is a problem though, because there is still some delay before it is presented. I went back and forth between not scaling the stabilization time or this approach, but this seems like the better route.

I also took this opportunity to speed up the test execution time by a couple seconds by dropping 2 second `longPress` actions to be 1s instead. This wasn't necessary, but I was running some tests that used long presses a lot and it seemed like a nice small improvement.

You can review each of these changes in the individual commits, which I do not plan to squash on merging this PR.

### DIFF
--- a/Sources/KIF/Additions/CALayer-KIFAdditions.m
+++ b/Sources/KIF/Additions/CALayer-KIFAdditions.m
@@ -41,12 +41,32 @@
 
             [layer.animationKeys enumerateObjectsUsingBlock:^(NSString *animationKey, NSUInteger idx, BOOL *innerStop) {
                 CAAnimation *animation = [layer animationForKey:animationKey];
-                double beginTime = [animation beginTime];
+
                 double completionTime = [animation KIF_completionTime];
 
-                // Ignore long running animations (> 1 minute duration)
-                if (currentTime >= beginTime && completionTime < currentTime + 60 && currentTime < completionTime) {
+                // Ignore long running animations (> 1 minute duration), as we don't want to wait on them
+                if (completionTime > currentTime + 60) {
+                    return;
+                }
+
+                // If an animation is set to be removed on completion, it must still be in progress if we enumerated it
+                // This is the default behavior for animations, so we should often hit this codepath.
+                if ([animation isRemovedOnCompletion]) {
                     result = YES;
+                } else if ([animation.delegate isKindOfClass:NSClassFromString(@"UIViewAnimationState")]) {
+                    // Use a private property on the private class to determine if the animation state has completed
+                    BOOL animationDidStopSent = [[(NSObject *)animation.delegate valueForKey:@"_animationDidStopSent"] boolValue];
+
+                    if (!animationDidStopSent) {
+                        result = YES;
+                    }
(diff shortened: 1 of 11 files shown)


Sources/KIF/Additions/CALayer-KIFAdditions.m
```diff
@@ -41,12 +41,32 @@ - (BOOL)hasAnimations
...
+                } else if ([animation.delegate isKindOfClass:NSClassFromString(@"UIViewAnimationState")]) {
+                    // Use a private property on the private class to determine if the animation state has completed
+                    BOOL animationDidStopSent = [[(NSObject *)animation.delegate valueForKey:@"_animationDidStopSent"] boolValue];
+
+                    if (!animationDidStopSent) {
+                        result = YES;
+                    }
+                } else if (currentTime > completionTime) {
+                    // Otherwise, use the completion time to determine if the animation has been completed.
+                    // This doesn't seem to always be exactly right however.
```
RoyalPineapple, line 65: Curious about the cases where it's not quite right, can you elaborate?

RoyalPineapple (approved):
Great work, thanks for taking this on.

Sources/KIF/Additions/CALayer-KIFAdditions.m
justinseanmartin: What I was seeing was when running `TypingTests.testEnteringTextIntoFirstResponder`, it would try to tap on "Select All" before it was finished being presented on Xcode 16. For some reason, startTime + duration was before the animation completed if I was watching it run in the simulator.

congt (approved):

PR #1307, opened 2024-09-24
Fix UICollectionView cell enumeration hitting asserts in Xcode 16

### LINKED ISSUE #1303: Broken tests with Xcode 16-RC / iOS 18
We are experiencing failures in tests written with `KIF`, and we aren't sure if it's related to the framework and the new versions of Xcode & iOS, or it is a more general problem.

In our project we have both `UITests` running with `XCUITest` and `KIF`. However the first ones are working as they were with `Xcode 15.4`, but not so much when running KIF-based ones.

The most common error we're finding is the following:
```
KIFTestActor.m:89: Expected dequeued view to be returned to the collection view in preparation for display. When the collection view's data source is asked to provide a view for a given index path, ensure that a single view is dequeued and returned to the collection view. Avoid dequeuing views without a request from the collection view. For retrieving an existing view in the collection view, use -[UICollectionView cellForItemAtIndexPath:] or -[UICollectionView supplementaryViewForElementKind:atIndexPath:]. Dequeued view: <XXXModule.XXXValuePropositionCell: 0x15fa73180; baseClass = UICollectionViewCell; frame = (383 0; 375 415); layer = <CALayer: 0x600004397bc0>>; Collection view: <UICollectionView: 0x1211e8200; frame = (0 0; 375 415); clipsToBounds = YES; gestureRecognizers = <NSArray: 0x6000011effc0>; backgroundColor = UIExtendedGrayColorSpace 0 0; layer = <CALayer: 0x6000044eb020>; contentOffset: {0, 0}; contentSize: {375, 415}; adjustedContentInset: {0, 0, 0, 0}; layout: <UICollectionViewCompositionalLayout: 0x1225a70f0>; dataSource: <XXXModule.XXXValuePropositionsCarouselView: 0x12259df10; frame = (0 140; 375 423); layer = <CALayer: 0x6000044eb120>>> (NSInternalInconsistencyException)
```

From our initial investigations, we have some screens in which we have a basic implementation, returning just one cell type, which is registered, while in some others we have a guard clause returning `UICollectionViewTableViewCell()` in the else clause, in which we have placed a breakpoint and wasn't hit while running tests and failing.

Furthermore, not running the app in a device, nor running `XCUITest` based tests we experience similar failures. Similarly, if I run `KIF tests` targeting `iOS 17.x` in the same `Xcode 16-RC` they don't fail, so the problem seems to be related to iOS 18.

Any help or guidance would be much appreciated.

justinseanmartin:
In Xcode 16, an assertion was to UICollectionView:

> Expected dequeued view to be returned to the collection view in preparation for display. When the collection view's data source is asked to provide a view for a given index path, ensure that a single view is dequeued and returned to the collection view. Avoid dequeuing views without a request from the collection view. For retrieving an existing view in the collection view, use -[UICollectionView cellForItemAtIndexPath:] or -[UICollectionView supplementaryViewForElementKind:atIndexPath:]. 

Rather than calling `-[UICollectionViewDataSource collectionView:cellForItemAtIndexPath:]` and `[UICollectionViewDelegate collectionView:willDisplayCell:forItemAtIndexPath:]`, instead we will scroll the collection view's content offset to make the offscreen cells be realized by UIKit. If the cell matches the predicate, then we will return the matching element (which should now be on screen). If a match isn't found in any of the cells, scroll back to the original scroll position and keep searching the view hierarchy for a matching element.

One notable thing to call out is that unlike cell enumeration in a UITableView using `cellForRowAtIndexPath:`, the UICollectionView equivalent `cellForItemAtIndexPath:` will return a nil cell unless a layour pass is performed. We could either spin the runloop and let UIKit update on its own or directly call `layoutIfNeeded` on the collection view. We're going with the latter, because it is faster on Xcode 16 and avoids a bunch of unnecessary and distracting redrawing of the screen at each scroll position.

A new API `[viewTester usingCurrentFrame]` was introduced in #1296 that can avoid this logic of implicitly scrolling collection views where that is needed or desired.

Fixes #1303

CC: @jdev7 @danielbarela @ispatel22 - Could you give this change a test locally and see if this addresses the issues you're seeing and see if it surfaces any other problems? I've landed a couple other changes recently, so it is possible that those could also have altered behavior in minor ways (#1296, #1300 and #1301).

### DIFF
--- a/Sources/KIF/Additions/UIView-KIFAdditions.m
+++ b/Sources/KIF/Additions/UIView-KIFAdditions.m
@@ -17,6 +17,9 @@
 #import "KIFUITestActor.h"
 #import <WebKit/WebKit.h>
 
+#define DRAG_TOUCH_DELAY 0.01
+#define CELL_SCROLL_DELAY_STABILIZATION 0.05
+
 double KIFDegreesToRadians(double deg) {
     return (deg) / 180.0 * M_PI;
 }
@@ -310,7 +313,6 @@ NS_INLINE BOOL StringsMatchExceptLineBreaks(NSString *expected, NSString *actual
                 }
             }];
 
-            CFTimeInterval delay = 0.05;
             for (NSUInteger section = 0, numberOfSections = [tableView numberOfSections]; section < numberOfSections; section++) {
                 for (NSUInteger row = 0, numberOfRows = [tableView numberOfRowsInSection:section]; row < numberOfRows; row++) {
                     if (!self.window) {
@@ -328,7 +330,7 @@ NS_INLINE BOOL StringsMatchExceptLineBreaks(NSString *expected, NSString *actual
                     }
 
                     @autoreleasepool {
-                        // Scroll to the cell and wait for the animation to complete. Using animations here may not be optimal.
+                        // Scroll to the cell and wait for the animation to complete.
                         CGRect sectionRect = [tableView rectForRowAtIndexPath:indexPath];
                         [tableView scrollRectToVisible:sectionRect animated:NO];
 
@@ -342,24 +344,25 @@ NS_INLINE BOOL StringsMatchExceptLineBreaks(NSString *expected, NSString *actual
                     }
 
                     // Note: using KIFRunLoopRunInModeRelativeToAnimationSpeed here may cause tests to stall
-                    CFRunLoopRunInMode(UIApplicationCurrentRunMode, delay, false);
+                    CFRunLoopRunInMode(UIApplicationCurrentRunMode, CELL_SCROLL_DELAY_STABILIZATION, false);
                     return [self accessibilityElementMatchingBlock:matchBlock disableScroll:NO];
                 }
             }
 
 			//if we're in a picker (scrollView), let's make sure we set the position back to how it was last set.
-            if(scrollView != nil && scrollContentOffset.x != -1.0)
+            if (scrollView != nil && scrollContentOffset.x != -1.0)
             {
                 [scrollView setContentOffset:scrollContentOffset];
             } else {
                 [tableView setContentOffset:initialPosition.origin];
             }
-            CFRunLoopRunInMode(UIApplicationCurrentRunMode, delay, false);
+
+            CFRunLoopRunInMode(UIApplicationCurrentRunMode, CELL_SCROLL_DELAY_STABILIZATION, false);
         } else if ([self isKindOfClass:[UICollectionView class]]) {
             UICollectionView *collectionView = (UICollectionView *)self;
-            
+            CGRect initialPosition = CGRectMake(collectionView.contentOffset.x, collectionView.contentOffset.y, collectionView.frame.size.width, collectionView.frame.size.height);
             NSArray *indexPathsForVisibleItems = [collectionView indexPathsForVisibleItems];
-            
+
             for (NSUInteger section = 0, numberOfSections = [collectionView numberOfSections]; section < numberOfSections; section++) {
                 for (NSUInteger item = 0, numberOfItems = [collectionView numberOfItemsInSection:section]; item < numberOfItems; item++) {
                     if (!self.window) {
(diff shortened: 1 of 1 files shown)


danielbarela:
I tested your fix, and it appears to fix my problem related to the collection view.  Thank you.

congt (approved):

Sources/KIF/Additions/UIView-KIFAdditions.m
```diff
@@ -373,36 +376,26 @@ - (UIAccessibilityElement *)accessibilityElementMatchingBlock:(BOOL(^)(UIAccessi
...
-
-                        // The cell contents might change just prior to being displayed, so we simulate the cell appearing onscreen
-                        if ([collectionView.delegate respondsToSelector:@selector(collectionView:willDisplayCell:forItemAtIndexPath:)]) {
-                            [collectionView.delegate collectionView:collectionView willDisplayCell:cell forItemAtIndexPath:indexPath];
-                        }
-                        
+                        CGRect visibleRect = [collectionView layoutAttributesForItemAtIndexPath:indexPath].frame;
+                        [collectionView scrollRectToVisible:visibleRect animated:NO];
+                        [collectionView layoutIfNeeded];
+                        UICollectionViewCell *cell = [collectionView cellForItemAtIndexPath:indexPath];
```
kyleve, line 382: I don't know the KIF internal best practices, but is there some warning or error we should throw if this returns a nil cell? Or meh?

Sources/KIF/Additions/UIView-KIFAdditions.m
```diff
@@ -373,36 +376,26 @@ - (UIAccessibilityElement *)accessibilityElementMatchingBlock:(BOOL(^)(UIAccessi
...
                         UIAccessibilityElement *element = [cell accessibilityElementMatchingBlock:matchBlock notHidden:NO disableScroll:NO];
-
-                        // Remove the cell from the collection view so that it doesn't stick around
-                        [cell removeFromSuperview];
                         
                         // Skip this cell if it isn't the one we're looking for
-                        // Sometimes we get cells with no size here which can cause an endless loop, so we ignore those
```
justinseanmartin, line 390: I'm not sure why this would cause an endless loop, but I'm guessing that realizing these cells in a more realistic way may sidestep this issue. We can always add this back later if need be.

Sources/KIF/Additions/UIView-KIFAdditions.m
```diff
@@ -373,36 +376,26 @@ - (UIAccessibilityElement *)accessibilityElementMatchingBlock:(BOOL(^)(UIAccessi
...
                     // Note: using KIFRunLoopRunInModeRelativeToAnimationSpeed here may cause tests to stall
-                    CFRunLoopRunInMode(UIApplicationCurrentRunMode, 0.5, false);
-                    
-                    // Now try finding the element again
+                    CFRunLoopRunInMode(UIApplicationCurrentRunMode, CELL_SCROLL_DELAY_STABILIZATION, false);
                     return [self accessibilityElementMatchingBlock:matchBlock disableScroll:NO];
                 }
             }
+
+            [collectionView setContentOffset:initialPosition.origin];
```
kyleve, line 397: We do a `layoutIfNeeded` after `layoutIfNeeded` above; should we do that here too?

Sources/KIF/Additions/UIView-KIFAdditions.m
justinseanmartin: I could NSAssert that the cell is non-nil to abort the search. I'll add that and see how it goes. Seems to pass locally at least.

kyleve: Honestly, I don't know if an assert is worth it – maybe a log or something? Unclear on the risk of false positives vs. false negatives where the test would just fail.

Sources/KIF/Additions/UIView-KIFAdditions.m
justinseanmartin: Sure, doesn't really hurt. I'll add it for consistency.

jbmorgan (approved):

kyleve (approved):

PR #1308, opened 2024-10-03
[dnr] [dnm] removing assert

RoyalPineapple:
our project is hitting this assert a ton, im putting up this pr to test without a hard assert.


justinseanmartin:
Try https://github.com/justinseanmartin/KIF/commit/4ff16aa141f3a43cc631fb4577a241481cd46ca3 and see if that works better. That didn't fix all the places we were seeing the assert in the full test suite, but did fix most of them. You can also combine it with `usingCurrentFrame` API if you know you don't want to scroll.

PR #1319, opened 2025-01-23
Change the collection view changes to iOS 18

mjohnson12:
Only use the new scrolling logic when on iOS 18+

### DIFF
--- a/Scripts/ci.sh
+++ b/Scripts/ci.sh
@@ -15,7 +15,7 @@ fi
 rm -rf ${PWD}/build
 
 # Run KIF Tests
-env NSUnbufferedIO=YES xcodebuild test -project KIF.xcodeproj -scheme KIF -derivedDataPath=${PWD}/build/KIF -destination "platform=iOS Simulator,${SIMULATOR}" | xcpretty -c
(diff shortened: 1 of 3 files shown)


Scripts/ci.sh
```diff
@@ -15,7 +15,7 @@ fi
 rm -rf ${PWD}/build
 
 # Run KIF Tests
-env NSUnbufferedIO=YES xcodebuild test -project KIF.xcodeproj -scheme KIF -derivedDataPath=${PWD}/build/KIF -destination "platform=iOS Simulator,${SIMULATOR}" | xcpretty -c
+env NSUnbufferedIO=YES xcodebuild test -project KIF.xcodeproj -scheme KIF -derivedDataPath=${PWD}/build/KIF -destination "platform=iOS Simulator,${SIMULATOR}" CODE_SIGNING_ALLOWED=NO | xcbeautify
```
jbmorgan, line 18: What's the reason for adding `CODE_SIGNING_ALLOWED=NO` here?

congt (approved):

Scripts/ci.sh
mjohnson12: Because it was failing to build with a you need a signing cert/company chosen.
Tim was trying to fix it here: https://github.com/kif-framework/KIF/pull/1320

Issue #1302, opened 2024-09-09
SwiftUI button being tapped but the click is not registered in XCode 16 beta.


ispatel22:
UI elements of a SwiftUI view get tapped but are not registered. For example, if there is a button with an accessibility identifier, KIF detects and taps on it, but the button's action method is not getting called. This issue only happens if you run the project on the latest XCode 16 beta. This bug does NOT occur on XCode 15. Below is the link to a sample repo that reproduces the issue.

[KIFTapBug repo](https://github.com/ispatel22/KIFTapBug/tree/main)

danielbarela:
I would also add that running the tests on the iOS17 simulator from the Xcode 16 beta does work, however running the same test on the iOS18 simulator does not.

ispatel22:
After debugging, I noticed issues with gesture recognizer on the button in the sample repo.

**If the app is ran on iOS 18 Xcode 16:**
 - There is ONLY `UIHoverGestureRecognizer` on the SwiftUIView
I manually added a test `UITapGestureRecognizer` in the KIF test and the call back method gets called.

Things I tried to fix:
- Added a .onTapGesture to the button in `MySwiftUIView.swift` but it did not work.

 **If the app is ran on iOS 17 Xcode 16:**
  - There is a `UIGestureRecognizer` and `UIHoverGestureRecognizer` on the SwiftUIView and the test passes as intended.

**To see the issue:**
1. Put a break point after the following line ` UIView* showTextButton = [[viewTester usingIdentifier:@"ShowTextButtonIdentifier"] waitForView];` in `test_01`
2. Run the test and in the debugger, under `showTextButton` look for `gestureRecognizers`. Run the test for both iOS 18 and iOS 17

ispatel22:
@justinseanmartin @danielbarela Will this issue be fixed any time soon?

justinseanmartin:
I haven't had time to take a look just yet. If you'd like to dig in and see if you can find a root cause, I'd be happy to help with pointers and guidance and help get it merged. Even setting up an integration test in KIF integration suite would be helpful for having something to use to investigate the issue and verify the fix.

ispatel22:
@justinseanmartin I'm not sure what you mean by `we can diff the touch events being sent from real user input compared to what KIF is doing to synthesize them`.

BartlomiejWlodarczak:
[...]

PR #1323, opened 2025-02-06
iOS 18 touch injection hit testing fix for SwiftUI view hierarchies

BartlomiejWlodarczak:
This PR should fix the issue https://github.com/kif-framework/KIF/issues/1302

There were additional mechanisms provided by Apple for SwiftUI / UIKit gesture interoperability. They needed to implement it somehow, hence the presence of new structures `SwiftUI.UIKitGestureContainer` and `_UIHitTestContext`. I tried to mimic what's actually going on during "real" touches on iOS 18.

I had also needed to bump the deployment target to iOS 13, since I was planning on adding a basic SwiftUI testing screen to the tests host app.

@justinseanmartin   
The UIWebView related logic can be ditched fully in favor of WKWebView if you decide to merge this PR.
I made adjustments for the rest of deprecation warnings.


Package.swift
```diff
@@ -5,7 +5,7 @@ import PackageDescription
 let package = Package(
     name: "KIF",
     platforms: [
-        .iOS(.v11)
+        .iOS(.v13)
```
tmyrden, line 8: I've tested this locally and it seems to be working @BartlomiejWlodarczak.

This is the only troublesome line in the PR. In order to consume via spm, line 1's `swift-tools-version:5.0` needs to be updated as well. 5.0 didn't know of v13. Bumping it to 5.3 on my fork of your branch resolved it - I believe 5.2 would be the min for v13 to work.

BartlomiejWlodarczak: Good spot! Thank you. I raised it to 5.3

lolindrath (commented):
Looking really good! I ran our tests and almost all of them passed. `clearTextFromView` is still failing - do you think that's related to this issue? The error I see is:

`with label "Label name" is not tappable. It may be blocked by other views.`

BartlomiejWlodarczak:
@lolindrath I can't really tell as I don't know what is the view hierarchy and test's logic in that failing test you mentioned. Would you like to share some more details?

Meanwhile I verified that current implementations of tester methods for text manipulation are working with the SwiftUI TextField view. I also added related tests.

justinseanmartin:
Thank you very much for contributing this! Tagging in @congt who can also tag in some other two repo admins to help move this forward. I'm on sabbatical for a while and not actively doing work things. I'll give it a once over, but I no longer have access to run the Block internal test suite to make sure it doesn't have any other side effects.

Sources/KIF/Additions/UITouch-KIFAdditions.m
```diff
@@ -124,10 +123,62 @@ - (void)setIsFromEdge:(BOOL)isFromEdge
...
+            id hitTestContext = ((id (*)(id, SEL, CGPoint, CGFloat))objc_msgSend)(UIHitTestContextClass, contextWithPointAndRadiusSel, point, 0);
+            
+            if (hitTestContext) {
+                /*
+                 From observation - this can be either of following types:
+                    - UIView type (e.g. when using UIViewRepresentable inside SwiftUI)
+                    - specialized SwiftUI view compatible with UIView,
+                    - newly introduced structure SwiftUI.UIKitGestureContainer implementing UIResponder interface.
+                    What's important it seems it is compatible with setView:(UIView *) method.
+                 */
```
justinseanmartin, line 165: Great context, thank you for including this.

justinseanmartin (approved):
Overall, this looks like a really well done change. Thanks for taking the time and effort to add all the tests. It seems like it could alter the behavior even in UIKit views, is that right? It's fine as long as it doesn't break behavior there.

FYI - There is a note about the supported iOS versions in the README that looks like it needs updating. It's already stale though. I'd guess this change also warrants a minor version bump and release cut to signify the notable added support. It is probably worth mentioning the SwiftUI support in the README as well. All this probably can be handled by repo maintainers after merging this PR as part of cutting a release, but defer to @congt et al on that.

BartlomiejWlodarczak:
@justinseanmartin thank you for your feedback on this!

Answering your question - whether or not it alters the current behavior for UIKit views. When there are just UIKit views in the view hierarchy it'll iterate it trying to create the new hit test context structure, which won't happen as it is only created when SwiftUI or SwiftUI + UIKit views are involved in the hierarchy. So it'll end up returning same ol' `hitTestView` that was used until now.

When I initially started working on this last year I tried to avoid that loop heuristically e.g. when the initially returned `hitTestView` was of a class containing keywords like `SwiftUI` / `UIHostingView` etc. but I found some SwiftUI views with really obfuscated names, so this wouldn't work and was failing in some situations.

So far after running the entire test suite before and after those changes I feel that the performance impact is negligible.

justinseanmartin:
> So far after running the entire test suite before and after those changes I feel that the performance impact is negligible.

@BartlomiejWlodarczak - I'm not worried about the perf impact either FWIW. KIF spends a lot of cycles walking the view hierarchy and I don't think that `-[UITouch initAtPoint:inWindow:]` gets called all that often by comparison.

bcriscuolo:
@justinseanmartin based on this work, what's the possibility of moving this into a formal KIF tagged release? We're eager to consume this and move to our functional automation on iOS 18 simulators.

Thanks!

BartlomiejWlodarczak:
@congt Sorted it out together with some build warnings from the logs.

ispatel22:
@justinseanmartin @congt can you make a new release version with the merged changes?

congt:
Here is the PR to bump the version https://github.com/kif-framework/KIF/pull/1325

PR #1329, opened 2025-04-29
Draw Signature KIF

JordanDenning:
Add ability to draw signature on element with "Signature Pad" accessibility label

Used in this [PR](https://github.com/squareup/ios-register/pull/124522)

### DIFF
--- a/Sources/KIF/Classes/KIFUITestActor.m
+++ b/Sources/KIF/Classes/KIFUITestActor.m
@@ -399,6 +399,19 @@ static BOOL KIFUITestActorAnimationsEnabled = YES;
     }];
 }
 
+-(void)drawSignature;
+{
+    UIView *signaturePad = nil;
+    UIAccessibilityElement *element = nil;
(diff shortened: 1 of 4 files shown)


Sources/KIF/Classes/KIFUITestActor.m
```diff
@@ -399,6 +399,19 @@ - (void)swipeFromEdge:(UIRectEdge)edge
     }];
 }
 
+-(void)drawSignature;
+{
+    UIView *signaturePad = nil;
+    UIAccessibilityElement *element = nil;
+    [self waitForAccessibilityElement:&element view:&signaturePad withLabel:@"Signature Pad" value:nil traits:UIAccessibilityTraitAllowsDirectInteraction tappable:YES];
```
justinseanmartin, line 406: This feels like an app specific concern. This seems like something that should be added as a class extension within your tests bundle rather than something that needs to be added directly to KIF itself.

FYI - The PR linked is to an internal repo, not one that is publicly available.

JordanDenning: Do you have any examples of that? I was having trouble with SSH so used codespaces but see that didn't quite work out how I was hoping

PR #1332, opened 2025-07-03
Fix a bug when determining whether an existing animation has completed

amorde:
Fixes a bug in the logic for determining if a given layer has active animations.

When encountering an animation that has already completed (its completion time is in the past), we incorrectly determined it was _not_ completed. This would lead to `waitForAnimationsToFinish()` to continue running until it hits the animation wait timeout, which defaults to 2 seconds.

### DIFF
--- a/Sources/KIF/Additions/CALayer-KIFAdditions.m
+++ b/Sources/KIF/Additions/CALayer-KIFAdditions.m
@@ -60,7 +60,7 @@
                     if (!animationDidStopSent) {
                         result = YES;
                     }
-                } else if (currentTime > completionTime) {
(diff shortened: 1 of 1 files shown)


mjohnson12:
Unfortunately, there might be tests that break now that this is correct.

amorde:
> Unfortunately, there might be tests that break now that this is correct.

Yeah, I've been working with @congt to verify this and it appears to not be too bad! Seems doable. A few places that maybe need some more typing delays etc.

congt (approved):

nabs-m (approved):

PR #1334, opened 2025-09-14
Ensure per-phase UIEvent creation in KIF tap simulation

krayc425:
## Background

* On iOS 26, KIF’s current touch injection logic in `-tapAtPoint:` sometimes fails to trigger taps on non-UIControl views (e.g. `UILabel`, `UIView`).  
* The root cause is that KIF reuses the **same `UIEvent` instance** for multiple touch phases (`Began` and `Ended`).  
* In earlier iOS versions this worked, but starting with iOS 26, it seems like UIKit appears to enforce stricter validation of `UIEvent` snapshots.  
* From my local testing of my app, if the event does not match the current `UITouch` phase, the system may ignore the injected event, resulting in taps being dropped.  

## Changes

- Update `-tapAtPoint:` to **create a new `UIEvent` for each touch phase (`.began` and `.ended`)**.   
- Ensure each event is freshly generated from the updated `UITouch` state before sending.

### DIFF
--- a/Sources/KIF/Additions/UIView-KIFAdditions.m
+++ b/Sources/KIF/Additions/UIView-KIFAdditions.m
@@ -599,13 +599,12 @@ NS_INLINE BOOL StringsMatchExceptLineBreaks(NSString *expected, NSString *actual
     // Handle touches in the normal way for other views
     UITouch *touch = [[UITouch alloc] initAtPoint:point inView:self];
     [touch setPhaseAndUpdateTimestamp:UITouchPhaseBegan];
-    
-    UIEvent *event = [self eventWithTouch:touch];
+    UIEvent *beganEvent = [self eventWithTouch:touch];
+    [[UIApplication sharedApplication] kif_sendEvent:beganEvent];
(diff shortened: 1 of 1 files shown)


justinseanmartin (approved):
Seems like something that ought to be fairly benign on previous versions. I'll wait on CI results, but should be able to merge this if things looks good.

justinseanmartin:
Think that there are any other changes needed for iOS26 before cutting a release? It'd be nice to add CI support to validate it and see if anything else is broken. I might be able to get to that later this week.

krayc425:
@justinseanmartin I think the Liquid Glass in the play would break some UI element testings, for example the TabBar / Alert / NavigationBar would have different UX or class hierarchy now. I haven't tested those since my use case doesn't enable Liquid Glass yet, but I think that part is worth an audit and fixes.

justinseanmartin:
FYI - I cut a release for this fix. Feel free to start using `v3.12.2` in place of using `master` branch or a specific sha.

